PR #38, opened 2022-07-23
Ingress node firewall configuration resource

msherif1234:
- configuration controller
- bindata daemonset manifest
- render and apply packages are based off sriov operator implementation
Unit-test
```
 k get ds -n ingress-node-firewall-system
NAME                           DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR                     AGE
ingress-node-firewall-daemon   0         0    
```


controllers/ingressnodefirewallconfig_controller.go
```diff
@@ -0,0 +1,137 @@
...
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.12.1/pkg/reconcile
+func (r *IngressNodeFirewallConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = context.Background()
+	logger := r.Log.WithValues("ingress node firewall config", req.NamespacedName)
+
+	instance := &ingressnodefwv1alpha1.IngressNodeFirewallConfig{}
+	err := r.Get(context.TODO(), req.NamespacedName, instance)
```
andreaskaris, line 67: use ctx instead of context.TODO()

controllers/ingressnodefirewallconfig_controller.go
```diff
@@ -0,0 +1,137 @@
...
+				ds.Spec.Template.Spec.Tolerations = config.Spec.Tolerations
+			}
+			err = scheme.Convert(ds, obj, nil)
+			if err != nil {
+				logger.Error(err, "Fail to convert DaemonSet to IngressNodeFirewallConfig object")
+				return err
+			}
+		}
+
+		if err := apply.ApplyObject(context.TODO(), r.Client, obj); err != nil {
```
andreaskaris, line 132: The same here: ctx instead of context.TODO()?

controllers/ingressnodefirewallconfig_controller.go
```diff
@@ -0,0 +1,137 @@
...
+	if err := r.syncIngressNodeFwConfigResources(instance); err != nil {
+		return ctrl.Result{}, nil
+	}
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *IngressNodeFirewallConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&ingressnodefwv1alpha1.IngressNodeFirewallConfig{}).
```
andreaskaris, line 93: add Owns like they do here: https://github.com/operator-framework/operator-sdk/blob/latest/testdata/go/v3/memcached-operator/controllers/memcached_controller.go#L218

And reconcile the DaemonSet, for example when a user hacks the DaemonSet image, or changes something else to the DaemonSet configuration, or deletes the DaemonSet, then this controller with `Owns` will take corrective action

controllers/ingressnodefirewallconfig_controller.go
```diff
@@ -0,0 +1,137 @@
...
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return ctrl.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return ctrl.Result{}, err
+	}
+
+	if req.Name != defaultIngressNodeFirwallCrName {
```
andreaskaris, line 79: type Firwall -> Firewall

controllers/ingressnodefirewallconfig_controller.go
```diff
@@ -0,0 +1,137 @@
...
+	IngressNodeFirewallManifestPath = "./bindata/manifests/daemon"
+)
+
+var ManifestPath = IngressNodeFirewallManifestPath
+
+// IngressNodeFirewallConfigReconciler reconciles a IngressNodeFirewallConfig object
+type IngressNodeFirewallConfigReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+	Log    logr.Logger
```
andreaskaris, line 46: possibly lower case, or does this need to be exported?

pkg/apply/apply.go
```diff
@@ -0,0 +1,59 @@
...
+	return existing, objDesc, err
+}
+
+// ApplyObject applies the desired object against the apiserver,
+// merging it with any existing objects if already present.
+func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstructured) error {
+	existing, objDesc, err := getExistingObject(ctx, client, obj)
+
+	if err != nil && apierrors.IsNotFound(err) {
+		log.Printf("does not exist, creating %s", objDesc)
```
andreaskaris, line 41: can you pass the logger from the reconciler in here and log to that one?

pkg/apply/apply.go
```diff
@@ -0,0 +1,59 @@
...
+func getExistingObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstructured) (*uns.Unstructured, string, error) {
+	name := obj.GetName()
+	namespace := obj.GetNamespace()
+	if name == "" {
+		return nil, "", errors.Errorf("Object %s has no name", obj.GroupVersionKind().String())
+	}
+	gvk := obj.GroupVersionKind()
+	// used for logging and errors
+	objDesc := fmt.Sprintf("(%s) %s/%s", gvk.String(), namespace, name)
+	log.Printf("reconciling %s", objDesc)
```
andreaskaris, line 26: can you pass the logger from the reconciler in here and log to that one?

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -0,0 +1,49 @@
...
+    metadata:
+      labels:
+        app: ingress-node-firewall-daemon
+        component: network
+        type: infra
+    spec:
+      containers:
+        - command: ['/bin/sh', '-c', 'mount bpffs /sys/fs/bpf -t bpf && daemon']
+          image: '{{.Image}}'
+          name: daemon
```
andreaskaris, line 21: Inject the node name into the container's environment variable:
~~~
      env:
        - name: NODE_NAME
          valueFrom:
            fieldRef:
              fieldPath: spec.nodeName
~~~

controllers/ingressnodefirewallconfig_controller.go
```diff
@@ -0,0 +1,137 @@
...
+		For(&ingressnodefwv1alpha1.IngressNodeFirewallConfig{}).
+		Complete(r)
+}
+
+func (r *IngressNodeFirewallConfigReconciler) syncIngressNodeFwConfigResources(config *ingressnodefwv1alpha1.IngressNodeFirewallConfig) error {
+	logger := r.Log.WithName("syncIngressNodeFirewallConfigResources")
+	logger.Info("Start")
+	data := render.MakeRenderData()
+
+	data.Data["Image"] = os.Getenv("DAEMONSET_IMAGE")
```
andreaskaris, line 102: Instead of taking DAEMONSET_IMAGE from env, shouldn't it come from config.Spec.Image?

controllers/ingressnodefirewallconfig_controller.go
```diff
@@ -0,0 +1,137 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if req.Name != defaultIngressNodeFirwallCrName {
+		logger.Error(err, "Invalid IngressNode firewall config resource name", "name", req.Name)
+		return ctrl.Result{}, nil // Return success to avoid requeue
+	}
+
+	if err := r.syncIngressNodeFwConfigResources(instance); err != nil {
+		return ctrl.Result{}, nil
```
andreaskaris, line 85: Add:
~~~
logger.Error(err, "Could not sync ingress node firewall config resource")
~~~

controllers/ingressnodefirewallconfig_controller.go
andreaskaris: Move apply.ApplyObject into the previous ifcondition (line 130), otherwise you apply non-deamonsets or null objects from` ./bindata/manifests/daemon/daemonset.yaml `

------

When you define, then the file parses into 2 objects (one daemonset, one null object):
~~~
[akaris@linux ingress-node-firewall (daemon)]$ cat ./bindata/manifests/daemon/daemonset.yaml 

---
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: ingr
~~~

You run apply.ApplyObject regardless of if this matches:
~~~
		if obj.GetKind() == "DaemonSet" &&
			(len(config.Spec.NodeSelector) > 0 || len(config.Spec.Tolerations) > 0) {
~~~

andreaskaris: Even after correcting that, I get:
~~~
"could not apply (apps/v1, Kind=DaemonSet) ingress-node-firewall-daemon: could not update object (apps/v1, Kind=DaemonSet) /ingress-node-firewall-daemon: an empty namespace may not be set when a resource name is provided",
~~~

api/v1alpha1/ingressnodefirewallconfig_types.go
```diff
@@ -0,0 +1,70 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// IngressNodeFirewallConfigSpec defines the desired state of IngressNodeFirewallConfig
+type IngressNodeFirewallConfigSpec struct {
+	// Ingress node firewall DaemonSet image
```
andreaskaris, line 29: Add:
`DaemonSetNamespace` here? So that you can define and parse the namespace where the DaemonSet should run?

controllers/ingressnodefirewallconfig_controller.go
```diff
@@ -0,0 +1,137 @@
...
+	data.Data["Image"] = os.Getenv("DAEMONSET_IMAGE")
+	objs, err := render.RenderDir(ManifestPath, &data)
+	if err != nil {
+		logger.Error(err, "Fail to render config daemon manifests")
+		return err
+	}
+
+	for _, obj := range objs {
+		if obj.GetKind() == "DaemonSet" &&
+			(len(config.Spec.NodeSelector) > 0 || len(config.Spec.Tolerations) > 0) {
```
andreaskaris, line 111: I think this here isn't needed and I'd remove this line:
If both NodeSelector and Tolerations are empty, then the default taint for masters is in place, and the DS will spawn the pods simply on all worker nodes.

pkg/apply/apply.go
```diff
@@ -0,0 +1,59 @@
...
+	existing.SetGroupVersionKind(gvk)
+	err := client.Get(ctx, types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}, existing)
+	return existing, objDesc, err
+}
+
+// ApplyObject applies the desired object against the apiserver,
+// merging it with any existing objects if already present.
+func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstructured) error {
+	existing, objDesc, err := getExistingObject(ctx, client, obj)
+
```
andreaskaris, line 39: Together with the `Owns` method, you must set:
https://github.com/operator-framework/operator-sdk/blob/latest/testdata/go/v3/memcached-operator/controllers/memcached_controller.go#L195
~~~
	// Set Memcached instance as the owner and controller
	ctrl.SetControllerReference(<owner configuration>, <daemonset>, r.Scheme)
~~~
That way, the ownerref is added to the daemonset and the ds is linked to the configuration object. Whenever the daemonset is messed with, the reconciler is triggered immediately.

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -0,0 +1,49 @@
...
+          volumeMounts:
+            - name: bpf-maps
+              mountPath: /sys/fs/bpf
+              mountPropagation: Bidirectional
+      nodeSelector:
+        kubernetes.io/os: linux
+      serviceAccountName: daemon
+      shareProcessNamespace: true
+      terminationGracePeriodSeconds: 2
+      tolerations:
```
andreaskaris, line 38: Remove that toleration here, you're setting it programmatically via the configuration crd

controllers/ingressnodefirewallconfig_controller.go
```diff
@@ -0,0 +1,137 @@
...
+				logger.Error(err, "Fail to convert IngressNodeFirewallConfig object to DaemonSet")
+				return err
+			}
+			if len(config.Spec.NodeSelector) > 0 {
+				ds.Spec.Template.Spec.NodeSelector = config.Spec.NodeSelector
+			}
+			if len(config.Spec.Tolerations) > 0 {
+				ds.Spec.Template.Spec.Tolerations = config.Spec.Tolerations
+			}
+			err = scheme.Convert(ds, obj, nil)
```
andreaskaris, line 125: Right above this line, add:
~~~
ctrl.SetControllerReference(config, ds, r.Scheme)
~~~

This will set the owner reference to the DS and it will allow us to monitor that DS's updates, deletes, etc:
~~~
    ownerReferences:
    - apiVersion: ingress-nodefw.ingress-nodefw/v1alpha1
      blockOwnerDeletion: true
      controller: true
      kind: IngressNodeFirewallConfig
      name: ingressnodefirewallconfig
      uid: b5c6875e-7ae0-4daa-8cea-fae191ff14f7
~~~

andreaskaris: And then with those 2 changes (Owns and the owner reference), the DS is now immediately recreated or updated when something happens to it:
~~~
[akaris@linux ingress-node-firewall (daemon)]$ oc get ds
NAME                           DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
ingress-node-firewall-daemon   3         3         3       3            3           kubernetes.io/os=linux   7m25s
[akaris@linux ingress-node-firewall (daemon)]$ oc delete ds ingress-node-firewall-daemon
daemonset.apps "ingress-node-firewall-daemon" deleted
[akaris@linux ingress-node-firewall (daemon)]$ oc get ds
NAME                           DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
ingress-node-firewall-daemon   3         3         2       3            2           kubernetes.io/os=linux   4s
[akaris@linux ingress-node-firewall (daemon)]$ oc get ds
NAME                           DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
ingress-node-firewall-daemon   3         3         3       3            3           kubernetes.io/os=linux   7s
~~~

controllers/ingressnodefirewallconfig_controller.go
msherif1234: we aren't using the spec and we don't need it we will rely on envvar setting I will remove it from the CR

controllers/ingressnodefirewallconfig_controller.go
msherif1234: yeah sorry

controllers/ingressnodefirewallconfig_controller.go
msherif1234: true I do check below to avoid update template so I can remove that one

controllers/ingrnodefwconfig_controller.go
```diff
@@ -0,0 +1,147 @@
...
+	logger.Info("Start")
+	data := render.MakeRenderData()
+
+	data.Data["Image"] = os.Getenv("DAEMONSET_IMAGE")
+	data.Data["NameSpace"] = r.Namespace
+
+	objs, err := render.RenderDir(ManifestPath, &data)
+	if err != nil {
+		logger.Error(err, "Fail to render config daemon manifests")
+		return err
```
andreaskaris, line 114: I think the Reconcile loop already logs the failure when you return err. So this here seems to be redundant (would log the error here, then the returned error).

Perhaps better to just:
~~~
return fmt.Errorf("failed to render config daemon manifests, err: %q", err)
~~~

controllers/ingrnodefwconfig_controller.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	for _, obj := range objs {
+		if obj.GetKind() == "DaemonSet" {
+			scheme := kscheme.Scheme
+			ds := &appsv1.DaemonSet{}
+			err = scheme.Convert(obj, ds, nil)
+			if err != nil {
+				logger.Error(err, "Fail to convert IngressNodeFirewallConfig object to DaemonSet")
```
andreaskaris, line 123: Same here, logger.Error and then return err seem redundant to me

controllers/ingrnodefwconfig_controller.go
```diff
@@ -0,0 +1,147 @@
...
+			}
+			if len(config.Spec.Tolerations) > 0 {
+				ds.Spec.Template.Spec.Tolerations = config.Spec.Tolerations
+			}
+			if err := controllerutil.SetControllerReference(config, obj, r.Scheme); err != nil {
+				return errors.Wrapf(err, "Failed to set controller reference to %s %s", obj.GetNamespace(), obj.GetName())
+			}
+			err = scheme.Convert(ds, obj, nil)
+			if err != nil {
+				logger.Error(err, "Fail to convert DaemonSet to IngressNodeFirewallConfig object")
```
andreaskaris, line 137: the same nit: here about logging and then returning the same error

pkg/apply/apply.go
```diff
@@ -0,0 +1,63 @@
...
+		err := client.Create(ctx, obj)
+		if err != nil {
+			return err
+		}
+		log.Printf("successfully created %s", objDesc)
+		return nil
+	}
+
+	// Merge the desired object with what actually exists
+	if err := MergeObjectForUpdate(existing, obj); err != nil {
```
andreaskaris, line 50: After a quick glance, I'm wondering if merge.go is necessary or if we can get rid of it.

I suppose that `mergeMetadataForUpdate` is used so that you can do `equality.Semantic.DeepEqual` --- but you could also just DeepEqual the specs and status, couldn't you?

I know that this is generic code, but it's for Deployment, ServiceAccount and Service. All things that we are not going to manipulate. It seems to me that the entire merge part might be redundant?

But I very likely am missing something :-)

controllers/ingrnodefwconfig_controller.go
```diff
@@ -0,0 +1,147 @@
...
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&ingressnodefwv1alpha1.IngrNodeFwConfig{}).
+		Owns(&appsv1.DaemonSet{}).
+		Complete(r)
+}
+
+func (r *IngrNodeFwConfigReconciler) syncIngressNodeFwConfigResources(ctx context.Context, config *ingressnodefwv1alpha1.IngrNodeFwConfig) error {
+	logger := r.Log.WithName("syncIngressNodeFirewallConfigResources")
+	logger.Info("Start")
+	data := render.MakeRenderData()
```
andreaskaris, line 106: I know this is currently way more generic, but we could also just fopen the single daemon.yaml file that we're interested in, read it into memory, and then use "k8s.io/apimachinery/pkg/util/yaml" or "gopkg.in/yaml.v2" to unmarshal the single DaemonSet that we're interested in. That would cut down on code at the cost of being less generic, of course.

controllers/ingrnodefwconfig_controller.go
msherif1234: this is kind of more of less the operator-sdk they don't propagate the err string just the err itself so I iwll stick with that flow ok ?

pkg/apply/apply.go
msherif1234: the idea from merge it walks the existing objects and merge them to the new object so at the end we really just update the delta so you don't lose prev configured objects.
I agree with you not really needed for our ds deployement I will take it off and bring it back if needed

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -0,0 +1,43 @@
...
+                  fieldPath: spec.nodeName
+          image: '{{.Image}}'
+          name: daemon
+          securityContext:
+            privileged: true
+            runAsUser: 0
+            capabilities:
+              add:
+                - CAP_BPF
+                - CAP_NET_ADMIN
```
martinkennelly, line 32: can you add this as well so we can see the last 80 lines of the log when a pod errors out:
```yaml
terminationMessagePolicy: FallbackToLogsOnError
```

controllers/ingrnodefwconfig_controller.go
```diff
@@ -0,0 +1,147 @@
...
+// +kubebuilder:rbac:groups=apps,namespace=ingress-node-firewall-system,resources=daemonsets,verbs=get;list;watch;create;update;patch;delete
+
+//+kubebuilder:rbac:groups=ingress-nodefw.ingress-nodefw,resources=ingrnodefwconfigs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=ingress-nodefw.ingress-nodefw,resources=ingrnodefwconfigs/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=ingress-nodefw.ingress-nodefw,resources=ingrnodefwconfigs/finalizers,verbs=update
+// +kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=get;list;watch;create;update;patch;delete
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
```
martinkennelly, line 61: Remove everything from `TODO...` on

controllers/ingrnodefwconfig_controller.go
andreaskaris: sounds good

controllers/ingrnodefwconfig_controller_test.go
```diff
@@ -0,0 +1,133 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+	ingressnodefwv1alpha1 "ingress-node-firewall/api/v1alpha1"
```
martinkennelly, line 6: format

controllers/suite_test.go
```diff
@@ -17,7 +17,12 @@ limitations under the License.
 package controllers
 
 import (
+	"context"
```
martinkennelly, line 20: format imports

pkg/apply/apply.go
```diff
@@ -0,0 +1,58 @@
+package apply
+
+import (
+	"context"
+	"fmt"
+	"github.com/pkg/errors"
```
martinkennelly, line 6: format

martinkennelly:
@msherif1234 can you remove the BPF byte objects from this commit as well?

README.md
```diff
@@ -1,84 +1,36 @@
...
+
+controller-gen v0.9.0+
+
+## Usage
+Once the Ingress node firewall Operator is installed, you have to create a `IngrNodeFwConfig` custom resource to deploy an instance. The operator will consume this resource and create ingress node firewall daemonset `daemon` based on it. The `IngrNodeFwConfig` custom resource needs to be created inside the `ingress-node-firewall-system` namespace and be named `ingressnodefirewallconfig`. Only one `IngrNodeFwConfig` resource can exist in a cluster.
+
+Following is example of `IngrNodeFwConfig` resource:
+```yaml
+apiVersion: ingress-nodefw.ingress-nodefw/v1alpha1
+kind: IngrNodeFwConfig
```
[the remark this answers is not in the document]
msherif1234: NodeDeployment is too general and will collide with something else

api/v1alpha1/ingrnodefwconfig_types.go
```diff
@@ -0,0 +1,67 @@
...
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+
+// IngrNodeFwConfig is the Schema for the ingrnodefwconfigs API
+type IngrNodeFwConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   IngrNodeFwConfigSpec   `json:"spec,omitempty"`
```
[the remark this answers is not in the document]
msherif1234: operator-sdk is the one created and mark it like this not me this was part of the scafefolding

martinkennelly:
/lgtm

API renaming will be discussed later.

PR #39, opened 2022-07-23
Remove binaries from bin/ and fix .gitignore

andreaskaris:
That unfortunately won't prune the 60 mb from the repo, but at least it'll prevent that from happening in the future

Signed-off-by: Andreas Karis <ak.karis@gmail.com>

**- How to verify it**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,4 @@
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
-./bin/*
+bin/*


msherif1234:
You need to keep ebpf.o files

andreaskaris:
But wouldn't I do something like:
~~~
container: ebpf-generate
    buildah bud . -t <container name>
~~~

So every time I build the container, the ebpf object files a created at that moment? Isn't it bad practive to have binary files checked into a repo?

msherif1234:
> But wouldn't I do something like:
> 
> ```
> container: ebpf-generate
>     buildah bud . -t <container name>
> ```
> 
> So every time I build the container, the ebpf object files are created at that moment? Isn't it bad practice to have binary files checked into a repo?

All app using cilium they checked in generated .o so I didn’t feel bad including them we can open issue to remove them and explore options but later

PR #50, opened 2022-07-26
eBPF event logging with packet captures

msherif1234:
enable eBPF perf event map and add userspace code to process the events and decode the packet header

Example
```
2022/07/27 11:44:03 ========= Drop Event ===========
2022/07/27 11:44:03     ruleId 10 action Drop len 98
2022/07/27 11:44:03     ipv4 src addr 172.20.0.1 dst addr 172.20.0.2
2022/07/27 11:44:03     icmpv4 type 8 code 0
2022/07/27 11:44:17 ========= Drop Event ===========
2022/07/27 11:44:17     ruleId 20 action Drop len 74
2022/07/27 11:44:17     ipv4 src addr 172.20.0.1 dst addr 172.20.0.2
2022/07/27 11:44:17     tcp srcPort 43618 dstPort 8080
```


pkg/ebpf/ingress_node_firewall_events.go
```diff
@@ -11,10 +11,14 @@ import (
 	"syscall"
 
 	"github.com/cilium/ebpf/perf"
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
 )
 
+const maxPacketHeaderSize = 100
```
martinkennelly, line 18: is this also defined in C?

msherif1234: yeah the same, I found the binary.read() since it force the user to have fixed size while using 100 worked for ICMP it didn't work for TCP I need to find a better api ideally if it can deal with variable size

pkg/ebpf/ingress_node_firewall_events.go
```diff
@@ -57,12 +66,70 @@ func (infc *IngNodeFwController) ingressNodeFwEvents() error {
 				continue
 			}
 
+			// read the event header
+			if _, err := io.ReadFull(bytes.NewBuffer(record.RawSample), buf[:]); err != nil {
+				log.Printf("parsing perf event header err: %s", err)
```
martinkennelly, line 71: %s -> %v
Start print message with upper case

pkg/ebpf/ingress_node_firewall_events.go
```diff
@@ -23,7 +29,8 @@ func (infc *IngNodeFwController) ingressNodeFwEvents() error {
 	// described in the eBPF C program.
 	rd, err := perf.NewReader(objs.IngressNodeFirewallEventsMap, os.Getpagesize())
 	if err != nil {
-		return fmt.Errorf("creating perf event reader: %s", err)
+		log.Printf("failed creating perf event reader: %s", err)
```
martinkennelly, line 32: Start message with uppercase

pkg/ebpf/ingress_node_firewall_events.go
```diff
@@ -57,12 +66,70 @@ func (infc *IngNodeFwController) ingressNodeFwEvents() error {
...
+			// Note position of the bytes in the buf slice depends on the layout of bpfEventHdrSt struct
+			eventHdr.RuleId = binary.LittleEndian.Uint16(buf[2:4])
+			eventHdr.Action = buf[4]
+			eventHdr.PktLength = binary.LittleEndian.Uint16(buf[6:8])
+			packet := make([]byte, eventHdr.PktLength)
 			// Parse the perf event entry into a bpfEvent structure.
-			if err := binary.Read(bytes.NewBuffer(record.RawSample), binary.LittleEndian, &event); err != nil {
-				log.Printf("parsing perf event: %s", err)
+			if err := binary.Read(bytes.NewBuffer(record.RawSample[eventHdrSize:]), binary.LittleEndian, &packet); err != nil {
+				log.Printf("parsing perf event packet header : %s", err)
```
martinkennelly, line 81: %s -> %v

martinkennelly (commented):
/lgtm 
overall - just small nits.

pkg/ebpf/ingress_node_firewall_events.go
andreaskaris: nit - I'd go with `%q` as it will quote the error. otherwise, both %s and %v print the value.  +1 for starting upppercase
https://go.dev/play/p/-Dz3JOG28Va

pkg/ebpf/ingress_node_firewall_events.go
andreaskaris: ... or %q if we want it quoted?

pkg/ebpf/ingress_node_firewall_events.go
```diff
@@ -57,12 +66,70 @@ func (infc *IngNodeFwController) ingressNodeFwEvents() error {
...
 				continue
 			}
+			log.Printf("========= Drop Event ===========")
```
andreaskaris, line 84: Will this log every dropped packet? I think for now, that level of verbosity is fine, but for production environments, we should only log the first drop for a given flow or not at all

msherif1234: we can have it debug level or something we can't know 1st pkt from flow or not

controllers/ingressnodefirewall_controller_test.go
```diff
@@ -98,10 +97,10 @@ var _ = Describe("IngressNodeFirewall controller", func() {
 		Expect(k8sClient.DeleteAllOf(context.Background(), &infv1alpha1.IngressNodeFirewall{})).Should(Succeed())
 		Expect(k8sClient.DeleteAllOf(context.Background(), &v1.Node{})).Should(Succeed())
 		// Ignore any not found errors.
-		k8sClient.DeleteAllOf(
+		Expect(k8sClient.DeleteAllOf(
```
andreaskaris, line 100: That's was on purpose, though, see comment:
~~~
// Ignore any not found errors.
~~~

One of the tests delets all IngressNodeFirewallNodeState objects and thus DeleteAllOf will throw and error as it doesn't find any object. Did you rerun `make test` and it still passed? In that case, I'm o.k. and my assumption was wrong. But the Expect here yesterday made the test fail for me, so I removed it.

So perhaps _ := k8sClient.DeleteAllOf will make the linter happy while still simply ignoring the result here?

controllers/ingressnodefirewall_controller_test.go
```diff
@@ -98,10 +97,10 @@ var _ = Describe("IngressNodeFirewall controller", func() {
 		Expect(k8sClient.DeleteAllOf(context.Background(), &infv1alpha1.IngressNodeFirewall{})).Should(Succeed())
 		Expect(k8sClient.DeleteAllOf(context.Background(), &v1.Node{})).Should(Succeed())
 		// Ignore any not found errors.
```
andreaskaris, line 99: nit: if making the line below Expect().Should(),  then remove this comment.

PR #57, opened 2022-07-27
Move pinholes pkg and rename it to "failsaferules"

martinkennelly:


martinkennelly:
I needed to update the module name in order to import the allowrules pkg in the api pkg.

controllers/ingressnodefirewall_controller.go
```diff
@@ -18,11 +18,12 @@ package controllers
 
 import (
 	"context"
+
```
msherif1234, line 21: you need to change it PROJECT too

api/v1alpha1/ingressnodefirewall_webhook.go
```diff
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"net"
 
+	"github.com/openshift/ingress-node-firewall/pkg/allowrules"
```
msherif1234, line 23: should we call it failsaferules or something more generic ?

pkg/allowrules/allowrules.go
```diff
@@ -0,0 +1,73 @@
+package allowrules
+
+import (
+	"fmt"
+)
+
+type TransportProtoAllowRule struct {
+	serviceName string
```
msherif1234, line 8: this structure should contain the protocol

pkg/allowrules/allowrules.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func GetUDP() []AllowRules {
+	return udpAllowRules
+}
+
+type AllowRules interface {
+	//IsConflictTransport checks to see if a range or a single port conflicts with the allow rule. If endPort is 0, we assume
+	// only a single port and not a range.
+	IsConflictTransport(startPort, endPort uint16) (bool, error)
```
msherif1234, line 49: I don't think this belong to this pkg
also can u add method to get protocol and the port

pkg/allowrules/allowrules.go
martinkennelly: you can pull the rules for your particular protocol so theres no need to include it in the struct.
e.g.
```GetTCP()``` or ```GetUDP()```

File: pkg/allowrules/allowrules.go (hunk 4 above)
Comment (martinkennelly) [reply pkg/allowrules/allowrules.go:49]: Yeah, removed the interface. Pointless.
No need to add get protocol because you can pull the fail safes by protocol so you know what protocol you are getting.

File: controllers/ingressnodefirewall_controller.go (hunk 1 above)
Comment (martinkennelly) [reply controllers/ingressnodefirewall_controller.go:21]: I changed the PATH. I hope thats sufficient..

File: api/v1alpha1/ingressnodefirewall_webhook.go [hunk 5]
```diff
@@ -204,21 +145,58 @@ func validateRule(rule IngressNodeFirewallProtocolRule, infRulesIndex, ruleIndex
...
+		}
+		if rule.ProtocolRule.IsRange() {
+			start, end, err = rule.ProtocolRule.GetRange()
+			if err != nil {
+				return false, "", fmt.Errorf("failed to get rule ports range: %v", err)
+			}
+			if isConflict, err = allowRule.IsConflictRange(start, end); err != nil {
+				return false, "", err
+			} else if isConflict {
+				return true, fmt.Sprintf("port range is in conflict with access to %s", allowRule.GetServiceName()), nil
```
msherif1234, line 185: why not return port range conflict as an error and not return string at all ?

pkg/failsaferules/failsaferules.go
```diff
@@ -0,0 +1,73 @@
...
+
+func GetTCP() []TransportProtoFailSafeRule {
+	return tcp
+}
+
+func GetUDP() []TransportProtoFailSafeRule {
+	return udp
+}
+
+func (t TransportProtoFailSafeRule) IsConflict(port uint16) (bool, error) {
```
msherif1234, line 46: didn't you agree on moving conflict functions below back to the webhook ?

api/v1alpha1/ingressnodefirewall_webhook.go
martinkennelly: I want to distinguish between an error and a reason for a conflict.

martinkennelly: If you want me to squash it into one return arg, I will. Dont feel strongly.

msherif1234: I just noticed its empty string in most cases and sometimes error comes as err or string I think will make things easier if we use one way to report validation error , agree ?

pkg/failsaferules/failsaferules.go
martinkennelly: I thought you meant removing the interface and not all the type methods.

api/v1alpha1/ingressnodefirewall_webhook.go
martinkennelly: ya. agreed.

pkg/failsaferules/failsaferules.go
msherif1234: this is common pkg for use by webhook and ds the conflict functions are only used by webhook so I think they better stay there , agree ?

martinkennelly: Ya, makes sense. Moving them.

PR #64, opened 2022-07-29
Add failSafe rules to ebpf map

msherif1234:
program failSafe rules to end of LPM map
sample logs for the api server hitting failSafe rule
```
 kube-apiserver-144001  [003] dNs3. 14917.741463: bpf_trace_printk: TCP/UDP/SCTP packet rule_dstPortStart 8000 rule_dstPortEnd 9000 pkt_dstPort 6443
 kube-apiserver-144001  [003] dNs3. 14917.741471: bpf_trace_printk: ruleInfo (protocol 6, Id 95, action 2)
```


Makefile
```diff
@@ -149,6 +149,7 @@ install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~
 .PHONY: uninstall
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
+	$(KUSTOMIZE) build config/daemon | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
```
andreaskaris, line 152: good catch

config/manager/kustomization.yaml
```diff
@@ -14,4 +14,4 @@ patchesStrategicMerge:
 - env.yaml
 images:
 - name: controller
-  newName: quay.io/akaris/controller
+  newName: quay.io/mmahmoud/controller
```
andreaskaris, line 17: That file constantly changes. Just for the future, we might want to check if we add it to gitignore or something

config/samples/ingressnodefirewall-demo-2.yaml
```diff
@@ -9,7 +9,7 @@ spec:
     do-node-ingress-firewall: 'true'
   ingress:
   - sourceCIDRs:
-       - 172.20.0.0/24
+       - 172.0.0.0/8
```
andreaskaris, line 12: nit: 172.16.0.0/12 to cover the entire type B block instead of public addresses?

controllers/ingressnodefirewallnodestate_controller.go
```diff
@@ -115,3 +122,38 @@ func (r *IngressNodeFirewallNodeStateReconciler) syncIngressNodeFirewallResource
 	}
 	return nil
 }
+
+// addFailSaferules appends failSafe rules to user configured one
+func addFailSaferules(rules *[]infv1alpha1.IngressNodeFirewallProtocolRule) error {
```
andreaskaris, line 127: How does this work with our trie longest prefix first match? Can't a user define some-source/32:tcp/6443 or some-source/32:tcp/22 and that will bypass the failsafe rules?

config/manager/kustomization.yaml
msherif1234: yeah we need to add that to the .gitignore or not commit it

controllers/ingressnodefirewallnodestate_controller.go
msherif1234: no u can't configure anything against those pinholes because of the webhook

api/v1alpha1/webhook_suite_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"github.com/openshift/ingress-node-firewall/pkg/failsaferules"
```
martinkennelly, line 23: format

bpf/ingress_node_firewall.h
```diff
@@ -44,10 +44,6 @@ struct ruleStatistics_st {
         __u64 packets;
         __u64 bytes;
     } deny_stats;
-    struct nomatch_stats_st {
```
martinkennelly, line 47: why remove this?

martinkennelly (commented):
can you update the webhook code to limit the number of allowed user rules to `MAX_INGRESS_RULES` ?
Also, at the top of the webhook code, I define a const for max ingress rules. Import it from failsaferules pkg.

msherif1234:
> can you update the webhook code to limit the number of allowed user rules to `MAX_INGRESS_RULES` ? Also, at the top of the webhook code, I define a const for max ingress rules. Import it from failsaferules pkg.

I already did that for webhook and webhook test did I miss other places ?

bpf/ingress_node_firewall.h
msherif1234: stats key ruleID now this action takes place only if there is no rule hit at all so I don't have rule ID to add those stats on
also I am using ruleId 0 to catch global no match drop /CIDR with new trace

martinkennelly:
> > can you update the webhook code to limit the number of allowed user rules to `MAX_INGRESS_RULES` ? Also, at the top of the webhook code, I define a const for max ingress rules. Import it from failsaferules pkg.
> 
> I already did that for webhook and webhook test did I miss other places ?

When I looked originally, I *thought* I didnt see what I referring to. I see it now though.

controllers/ingressnodefirewallnodestate_controller.go
andreaskaris: ah ok thanks

PR #66, opened 2022-07-29
Export statistics as metrics

martinkennelly:


martinkennelly:
Havent fully tested the metrics exporting yet and I also notice the rbac proxies dont seem to be getting their certs created.

martinkennelly:
Removing the WIP incase anyone wants to help push this on over the next 36 hours.

martinkennelly:
I notice a test is failing also:
```
Summarizing 1 Failure:

[Fail] Ingress nodefirewall config Controller syncIngressNodeFwConfig [It] Should create manifests with images and namespace overriden 
/home/mkennell/repos/ingress-node-firewall/controllers/ingrnodefwconfig_controller_test.go:54
```

msherif1234:
> I notice a test is failing also:
> 
> ```
> Summarizing 1 Failure:
> 
> [Fail] Ingress nodefirewall config Controller syncIngressNodeFwConfig [It] Should create manifests with images and namespace overriden 
> /home/mkennell/repos/ingress-node-firewall/controllers/ingrnodefwconfig_controller_test.go:54
> ```

The error because you changed container name is manifest but not in Config unit test

msherif1234:
Why are you updating statistics? Kerenel code update the statistics we just need to read and report?

msherif1234:
ebpf2go structure can’t be exported that is why access them need to in ebpf package

martinkennelly:
> Why are you updating statistics? Kerenel code update the statistics we just need to read, aggregate and report?

That is what is happening in pkg metrics/statistics.go.

pkg/metrics/statistics.go
```diff
@@ -0,0 +1,163 @@
...
+					}
+
+					if result, ok = addUInt64Float64(stat.DenyStats.Packets, denyPackets); !ok {
+						log.Println("Overflow occurred during addition of deny packet statistic")
+					} else {
+						denyPackets = result
+					}
+
+					if result, ok = addUInt64Float64(stat.DenyStats.Bytes, denyBytes); !ok {
+						log.Println("Overflow occurred during addition of deny byte statistic")
```
msherif1234, line 135: why not handle the overflow by having a carry counter for each one and once detect overflow increment it?

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -49,9 +49,63 @@ spec:
             - name: bpf-maps
               mountPath: /sys/fs/bpf
               mountPropagation: Bidirectional
+        - name: kube-rbac-proxy
+          image: quay.io/openshift/origin-kube-rbac-proxy:latest
```
msherif1234, line 53: this is need to be a variable that we can render like DaemonSet image and we can default to quay in env.yaml similar to how we do DS image

msherif1234 (commented):
I reviewed stats logic

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -49,9 +49,63 @@ spec:
...
+              echo $(date -Iseconds) INFO: ingress-node-firewall-daemon-metrics-certs mounted, starting kube-rbac-proxy
+              exec /usr/bin/kube-rbac-proxy \
+                --logtostderr \
+                --secure-listen-address=:9301 \
+                --tls-cipher-suites=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256 \
+                --upstream=http://127.0.0.1:39301 / \
+                --tls-private-key-file=${TLS_PK} \
+                --tls-cert-file=${TLS_CERT}
+          ports:
+            - containerPort: 9301
```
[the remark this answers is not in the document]
martinkennelly: yes, its currently not reserved.

cmd/daemon.go
```diff
@@ -51,8 +52,9 @@ func init() {
 func main() {
 	var metricsAddr string
 	var probeAddr string
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
-	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	// We are host networked, we set default to loopback by default
+	flag.StringVar(&probeAddr, "health-probe-bind-address", "127.0.0.1:39300", "The address the probe endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "127.0.0.1:39301", "The address the metric endpoint binds to.")
```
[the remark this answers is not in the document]
martinkennelly: I wanted to be explicit showing that we do not listen on all interfaces but only loopback. We dont want to expose this port outside the node. I figured `:$PORT_NUM` would listen on all interfaces and thats not what I want even as a default for this component because we always have the rbac proxy.

> in case both pods are on the same node right ?

yes and also, we cannot reserve within OCP the same port for two components.

controllers/ingressnodefirewallnodestate_controller.go
```diff
@@ -116,6 +120,8 @@ func (r *IngressNodeFirewallNodeStateReconciler) syncIngressNodeFirewallResource
 		}
 	}
 
+	r.Stats.StartPoll(c.GetStatisticsMap(), time.Second)
```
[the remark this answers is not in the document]
martinkennelly: Good point. Ill do this! Thanks!

msherif1234:
we should also look at statistics update in e2e functional test as well agree ?

pkg/metrics/statistics.go
```diff
@@ -0,0 +1,163 @@
...
+	Namespace: metricINFNamespace,
+	Subsystem: metricINFSubsystemNode,
+	Name:      "packet_allow_total",
+	Help:      "The number of packets which results in an allow IP packet result",
+})
+
+var metricAllowPacketBytes = prometheus.NewGauge(prometheus.GaugeOpts{
+	Namespace: metricINFNamespace,
+	Subsystem: metricINFSubsystemNode,
+	Name:      "packet_allow_bytes",
```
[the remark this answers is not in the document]
martinkennelly: Thats not recommended. See here: https://github.com/prometheus/docs/blob/main/content/docs/practices/naming.md

pkg/metrics/statistics.go
```diff
@@ -0,0 +1,163 @@
...
+	Name:      "packet_deny_total",
+	Help:      "The number of packets which results in a deny IP packet result",
+})
+
+var metricDenyPacketBytes = prometheus.NewGauge(prometheus.GaugeOpts{
+	Namespace: metricINFNamespace,
+	Subsystem: metricINFSubsystemNode,
+	Name:      "packet_deny_bytes",
+	Help:      "The number of bytes for packets which results in an deny IP packet result",
+})
```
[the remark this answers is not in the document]
martinkennelly: Not recommended, see here: https://github.com/prometheus/docs/blob/main/content/docs/practices/naming.md

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.20.0
 	github.com/pkg/errors v0.9.1
+	github.com/prometheus/client_golang v1.12.1
```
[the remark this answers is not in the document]
martinkennelly: Wasnt needed. It was already imported.

martinkennelly:
> we should also look at statistics update in e2e functional test as well agree ?

Its rather hard to do it because I am setting Prometheus gauges and there is no interface I can use to swap out these gauges with a dummy during testing. There is one for all the Prometheus vector types but not gauge.

pkg/metrics/statistics.go
```diff
@@ -0,0 +1,163 @@
...
+	"time"
+
+	nodefwloader "github.com/openshift/ingress-node-firewall/pkg/ebpf"
+
+	"github.com/cilium/ebpf"
+	"github.com/prometheus/client_golang/prometheus"
+	controllerruntimemetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var metricAllowPacketCount = prometheus.NewGauge(prometheus.GaugeOpts{
```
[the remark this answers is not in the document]
martinkennelly: I removed `packets` cause it was pointless.

pkg/metrics/statistics.go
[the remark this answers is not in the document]
msherif1234: how about ?
```
allow_packets
allow_bytes
deny_packets
deny_bytes
```

martinkennelly:
I was able to test the e2e code and ensure it fully works by deleting the existing e2e code during testing. 
Theres a known issue with deleting an INF obj that causes failure of daemon.

config/openshift/rbac.yaml
```diff
@@ -9,4 +9,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: ingress-node-firewall-daemon
-  namespace: ingress-node-firewall-system
+  namespace: openshift-ingress-node-firewall
```
msherif1234, line 12: was that intentional ?

config/manager/env.yaml
```diff
@@ -17,3 +17,5 @@ spec:
                   fieldPath: metadata.namespace
             - name: KUBE_RBAC_PROXY_IMAGE
               value: "quay.io/openshift/origin-kube-rbac-proxy:latest"
+            - name: POLL_PERIOD_SECONDS
```
msherif1234, line 20: this ds env var u don't need it here

cmd/daemon.go
```diff
@@ -88,7 +94,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	stats := metrics.NewStatistics()
+	stats, err := metrics.NewStatistics(pollPeriod)
```
msherif1234, line 97: should the pollPeriod be `time.Duration` instead of `string` ?

msherif1234 (commented):
you need to read KUBE_PROXY env var in main as well

msherif1234:
nit: last commit has pkgs changes too will be nice if u have it in its own commit

config/openshift/rbac.yaml
martinkennelly: yes, within OCP, we want our product to start with `openshift-`.

config/manager/env.yaml
martinkennelly: I thought in a previous comment you wanted to expose the stats poll people as a config option and the only existing way to do it is with an env var via controller since it generates the daemonset.

cmd/daemon.go
martinkennelly: I wanted to handle the parsing of an integer and error handling within `NewStatistics`. 
I think its fine but willing to change it if you want.

config/openshift/rbac.yaml
msherif1234: ok the ns is hardcoded in many place including e2e we will need to clean that up based on the env we are running on "runtime" I will look at next

config/manager/env.yaml
msherif1234: so this patched to manager process what we expose here is what we render inside the controllers other DS env var stays in the ds manifest like the events log file for example and if someone wanted to change they chnage the manifset and the controller will generate new ds with the required envvar, make sense ?

cmd/daemon.go
msherif1234: ur call in most of the application I saw this is normally used as duration so u can use `1s` or `1m` style I will leave it to you to decide

msherif1234:
pls update the README with the new make commands

Makefile
```diff
@@ -160,23 +160,29 @@ endif
...
-	$(KUSTOMIZE) build config/daemon | kubectl apply -f -
 
 .PHONY: uninstall
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
-	$(KUSTOMIZE) build config/daemon | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
-	kubectl delete -f config/daemon/namespace.yaml --ignore-not-found=$(ignore-not-found)
+
+.PHONY: deploy-kind
+deploy-kind: manifests kustomize install-cert-manager ## Deploy controller to the K8s cluster specified in ~/.kube/config.
```
msherif1234, line 169: pls update the comment stating this deploy on KIND while the other deploy on OCP

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -35,8 +37,8 @@ spec:
                   fieldPath: metadata.namespace
             - name: EVENT_LOGGING_FILE
               value: "/tmp/ingress_node_firewall_events.log"
-          image: '{{.Image}}'
-          name: daemon
+            - name: POLL_PERIOD_SECONDS
+              value: "{{ .PollPeriodSeconds }}"
```
msherif1234, line 41: I don't think we need to render this u define the default here and user can edit the manifest to change ?

config/rbac/service.yaml
```diff
@@ -0,0 +1,39 @@
...
+  ports:
+    - name: metrics
+      port: 9301
+      protocol: TCP
+      targetPort: https
+  selector:
+    app: ingress-node-firewall-daemon
+  clusterIP: None
+  publishNotReadyAddresses: true
+  sessionAffinity: None
```
msherif1234, line 38: can u call out the statistics ports in the readme with example of how to query them with curl for example ?

controllers/ingressnodefirewallconfig_controller.go
```diff
@@ -106,6 +106,8 @@ func (r *IngressNodeFirewallConfigReconciler) syncIngressNodeFwConfigResources(c
 
 	data.Data["Image"] = os.Getenv("DAEMONSET_IMAGE")
 	data.Data["NameSpace"] = r.Namespace
+	data.Data["RBACProxyImage"] = os.Getenv("KUBE_RBAC_PROXY_IMAGE")
+	data.Data["PollPeriodSeconds"] = os.Getenv("POLL_PERIOD_SECONDS")
```
msherif1234, line 110: this might be too much for Pollperiod what do u think ? just envvar in ds will be enough IMO

test/e2e/functional/tests/e2e.go
```diff
@@ -192,4 +198,180 @@ var _ = Describe("Ingress Node Firewall", func() {
...
+				}
+
+				totalEndpoints := 0
+				for _, slice := range endpointSliceList.Items {
+					totalEndpoints += len(slice.Subsets)
+				}
+				if totalEndpoints > 0 {
+					return true, nil
+				}
+				return false, nil
```
msherif1234, line 254: do we need to return err here to fail the test ?

test/e2e/pods/pods.go
```diff
@@ -0,0 +1,49 @@
...
+	testclient "github.com/openshift/ingress-node-firewall/test/e2e/client"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/remotecommand"
+)
+
+// ExecCommand runs command in the pod and returns buffer output
+// copied entirely from SRIOV Network Operator
+func ExecCommand(cs *testclient.ClientSet, pod *corev1.Pod, command ...string) (string, string, error) {
```
msherif1234, line 16: Nice that will save me from the trouble of trying to use kubectl exec Thanks for adding this !!

bindata/manifests/daemon/daemonset.yaml
martinkennelly: Agree.

test/e2e/functional/tests/e2e.go
martinkennelly: We want it to loop around again incase the endpoints havent updated yet.

test/e2e/pods/pods.go
martinkennelly: Thank the SRIOV network operator folks. I see nearly the exact same function in upstream k8 code so maybe I should thank them too ;)

martinkennelly:
@msherif1234 Thanks for the great review.

PR #79, opened 2022-08-05
IngressNodeFirewall controller - implement unit tests, change NodeState CRD and implement merge

andreaskaris:
IngressNodeFirewall controller - implement unit tests, change NodeState CRD and implement merge

Signed-off-by: Andreas Karis <ak.karis@gmail.com>

**- What this PR does and why is it needed**

**- Special notes for reviewers**

**- How to verify it**

**- Description for the changelog**

PR #80, opened 2022-08-06
make ingress node firewall rules interface specific

msherif1234:
we need to have flexiability to allow specific rules per interface(s) hence we need to make ifidx part of LPM key for example
if we have rules with sourceCIDR `A` applied to eth0 traffic from eth1 with matching CIDR won't incorrectly execute eth0's rules.
```
2022/08/06 00:29:45     ruleId 10 action Drop len 98 if eth0
2022/08/06 00:29:45     ipv4 src addr 172.18.0.1 dst addr 172.18.0.2
2022/08/06 00:29:45     icmpv4 type 8 code 0
```


pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -195,27 +198,30 @@ func (infc *IngNodeFwController) makeIngressFwRulesMap(ingFirewallConfig ingress
...
 	objs := infc.objs
 	for _, ifaceName := range ifacesName {
 		ifaceName := ifaceName
 		// Look up the network interface by name.
 		iface, err := net.InterfaceByName(ifaceName)
 		if err != nil {
-			return fmt.Errorf("lookup network iface %q: %s", ifaceName, err)
+			return ifIdlist, fmt.Errorf("lookup network iface %q: %s", ifaceName, err)
 		}
+		ifIdlist = append(ifIdlist, uint32(iface.Index))
```
andreaskaris, line 212: can you return map[string]uint32  with ifaceName -> uint32? That way, callers of IngressNodeFwAttach can know which returned index belongs to which interface

andreaskaris (commented):
That's really cool, so you found a way to make rules work per interface :-)

pkg/ebpf/ingress_node_firewall_loader.go
msherif1234: Will do

andreaskaris:
There's a problem here on unattach when only a single iface is removed. Imagine eth0 and eth1 were attached, then you call a detach only for eth1. That'll trigger cleanup which closes all objects, even for eth0.
A possible solution (among other things) is in:
https://github.com/openshift/ingress-node-firewall/pull/79/commits/ea9ed065f1f5a5744547790fc9ceb85c1e9d4452#diff-fb37587accd34f0659e8e5c56a3aa5809ee881d5bf6d1310ef3651df3663fc78R233

But I think that all of this will need some further thoughts ... atm, that sync stuff is really broken for me as soon as I try to test it for idempotency.

msherif1234:
> There's a problem here on unattach when only a single iface is removed. Imagine eth0 and eth1 were attached, then you call a detach only for eth1. That'll trigger cleanup which closes all objects, even for eth0. A possible solution (among other things) is in: [ea9ed06#diff-fb37587accd34f0659e8e5c56a3aa5809ee881d5bf6d1310ef3651df3663fc78R233](https://github.com/openshift/ingress-node-firewall/commit/ea9ed065f1f5a5744547790fc9ceb85c1e9d4452#diff-fb37587accd34f0659e8e5c56a3aa5809ee881d5bf6d1310ef3651df3663fc78R233)
> 
> But I think that all of this will need some further thoughts ... atm, that sync stuff is really broken for me as soon as I try to test it for idempotency.

we just need simply a reference count to know when to remove objs take alook now I think this will do the job

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -25,12 +26,15 @@ const (
...
+	mu sync.Mutex
 	// eBPF objs to create/update eBPF maps
 	objs BpfObjects
 	// eBPF interfaces attachment objects
-	links []link.Link
+	links map[string]link.Link
 	// eBPF pingPath
 	pinPath string
+	// refCnt how many interfces use this BpfObjects
+	refCnt int
```
andreaskaris, line 37: Instead of using refCnt, would it make sense to:
~~~
delete(ifaceName, infc.links)
if len(ifnc.links) <= 0 {
...
}
~~~

msherif1234: agree will remove the counter

andreaskaris (approved):
/lgtm

Issue #78, opened 2022-08-05
Webhook should be in a separate pkg from the API


martinkennelly:
Otherwise, folks importing our API, also get all the webhook dependencies.
All credit for noticing this goes to @JoelSpeed

msherif1234:
PR #81 fix this issue

PR #81, opened 2022-08-06
Move webhook to its own package

martinkennelly:
This will allow API consumers not to have to import our
webhook imports when they just want to consume the API.

/cc @msherif1234


msherif1234:
While you are here can add something for webhook in e2e

martinkennelly:
> While you are here can add something for webhook in e2e

Sure. Good idea.

msherif1234:
fix #78

PR #86, opened 2022-08-08
Add OCP dockerfile to build DaemonSet.

msherif1234:
Need to have dockefile for OCP to build Daemonset image by the release team


andreaskaris:
/lgtm

Wrt -mod=vendor:
~~~
go help build
	-mod mode
		module download mode to use: readonly, vendor, or mod.
		By default, if a vendor directory is present and the go version in go.mod
		is 1.14 or higher, the go command acts as if -mod=vendor were set.
~~~

PR #90, opened 2022-08-08
Handle XDP attach "device or resource busy" error

msherif1234:
- handle multiple attach to the same interface
- make sure to unpin link and remove pinPath once objs are deleted

### DIFF
--- a/pkg/ebpf/ingress_node_firewall_loader.go
+++ b/pkg/ebpf/ingress_node_firewall_loader.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"os"
 	"path"
+	"strings"
 	"syscall"
 
 	ingressnodefwiov1alpha1 "github.com/openshift/ingress-node-firewall/api/v1alpha1"
@@ -17,9 +18,10 @@ import (
 )
 
 const (
-	xdpDeny   = 1 // XDP_DROP value
-	xdpAllow  = 2 // XDP_PASS value
-	bpfFSPath = "/sys/fs/bpf"
+	xdpDeny     = 1 // XDP_DROP value
+	xdpAllow    = 2 // XDP_PASS value
+	bpfFSPath   = "/sys/fs/bpf"
+	xdpEBUSYErr = "device or resource busy"
 )
 
 // IngNodeFwController structure is the object hold controls for starting
@@ -220,6 +222,10 @@ func (infc *IngNodeFwController) IngressNodeFwAttach(ifacesName []string, isDele
 				Interface: iface.Index,
 			})
 			if err != nil {
+				if strings.Contains(err.Error(), xdpEBUSYErr) {
+					log.Printf("Interface %s is already attached", ifaceName)
+					return ifMap, nil
+				}
 				return ifMap, fmt.Errorf("could not attach XDP program: %s", err)
 			}
 			lPinDir := path.Join(infc.pinPath, ifaceName+"_link")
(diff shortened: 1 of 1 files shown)


pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -220,6 +220,12 @@ func (infc *IngNodeFwController) IngressNodeFwAttach(ifacesName []string, isDele
 				Interface: iface.Index,
 			})
 			if err != nil {
+				if errno, ok := err.(syscall.Errno); ok {
+					if errno == syscall.EBUSY {
```
andreaskaris, line 224: That doesn't work because cilium return a string type error:
~~~
2022/08/09 10:41:34 akaris ---------------- can't create link: device or resource busy *errors.errorString
~~~

Here /development/ingress-node-firewall/vendor/github.com/cilium/ebpf/link/link.go:
~~~
  178     fd, err := sys.LinkCreate(&attr)                                                                                    
  179     if err != nil {                                                                                                     
  180         return nil, fmt.Errorf("can't create link: %s", err)                                                            
  181     } 
~~~

andreaskaris: This here works though:
~~~
  223             if err != nil {                                                                                             
  224                 if strings.Contains(err.Error(), "device or resource busy") {                                           
  225                     log.Printf("Interface %s is already attached", ifaceName)                                           
  226                     return ifMap, nil                                                                                   
  227                 }                                                                                                       
  228                 return ifMap, fmt.Errorf("could not attach XDP program: %s", err)                                       
  229             }  
~~~

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -238,19 +244,25 @@ func (infc *IngNodeFwController) IngressNodeFwAttach(ifacesName []string, isDele
 	return ifMap, nil
 }
 
-// cleanup will delete all link objects for all interfaces and remove all the maps
+// cleanup will delete interface's eBPF objects.
 func (infc *IngNodeFwController) cleanup(ifName string) error {
 	l, ok := infc.links[ifName]
```
andreaskaris, line 249: I just realized that there's a problem with this here:
a) we must hold an instance of infc in memory for the links[] to be populated, otherwise we get: "failed to find Link object for interface"  (I can solve that by creating a long-living instance of infc)
b) while I can easily add a single infc the ebpfsyncer singleton's state in my PR, and while I can solve that issue then for the same process, if the process crashes and then restarts and calls a cleanup for the interface, this will still fail

a) is easy to fix, b) not so much

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -238,19 +244,25 @@ func (infc *IngNodeFwController) IngressNodeFwAttach(ifacesName []string, isDele
...
+	if err := l.Unpin(); err != nil {
+		return fmt.Errorf("failed to unpin link for %s err: %q", ifName, err)
+	}
 	if err := l.Close(); err != nil {
-		return err
+		return fmt.Errorf("failed to close and detach link %s err: %q", ifName, err)
 	}
 	delete(infc.links, ifName)
 	if len(infc.links) == 0 {
 		if err := infc.objs.Close(); err != nil {
```
andreaskaris, line 261: That close here creates an issue. When I attach all interfaces, then detach all interfaces -> Close()
A subsequent attach will then fail with an invalid file descriptor because the objs are closed. There's no way for a user of an instance of infc to know if the object was closed, and it doesn't reinitialise closed objects by itself.

pkg/ebpf/ingress_node_firewall_loader.go
msherif1234: yep that is what I did :)

pkg/ebpf/ingress_node_firewall_loader.go
msherif1234: so we should just never close objs nor pinDir and if daemon crashes we create new instance and go from there

pkg/ebpf/ingress_node_firewall_loader.go
msherif1234: yeah we need to keep infc instance to be able to handle process cleanup and call the new api `CleaneBPFObjs`

pkg/ebpf/ingress_node_firewall_loader.go
msherif1234: as discussed on slack new api to cleanup when ds go down

andreaskaris:
/lgtm
To merge this in. I'll see what I can do with this in my PR and I'll add anything that I think is missing

PR #94, opened 2022-08-10
change how we do envtest target and linit script to work from ci repo

msherif1234:
It is causing issues when run unit-test from
release CI we get the following error
GOBIN=/go/src/github.com/openshift/ingress-node-firewall/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.0
25
go: sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.0: cannot query module due to -mod=vendor
26
make: *** [/go/src/github.com/openshift/ingress-node-firewall/bin/controller-gen] Error 1

### DIFF
--- a/hack/lint.sh
+++ b/hack/lint.sh
@@ -4,5 +4,11 @@
 
 VERSION="v1.45.2"
 
-docker run --rm -v $(pwd):/app:z -w /app -e GO111MODULE=on golangci/golangci-lint:${VERSION} \
+if [[ $(which podman) && $(podman --version) ]]; then
+  command=podman
+else
+  command=docker
+fi
+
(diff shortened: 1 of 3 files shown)


msherif1234:
https://github.com/openshift/release/pull/30959

hack/lint.sh
```diff
@@ -4,5 +4,5 @@
 
 VERSION="v1.45.2"
 
-docker run --rm -v $(pwd):/app:z -w /app -e GO111MODULE=on golangci/golangci-lint:${VERSION} \
+podman run --rm -v $(pwd):/app:z -w /app -e GO111MODULE=on golangci/golangci-lint:${VERSION} \
```
andreaskaris, line 7: nit: perhaps add detection here if podman is present, if so, use it. Otherwise use docker

PR #96, opened 2022-08-10
Interfaces validation pkg

### LINKED ISSUE #91: Make sure to check valid interfaces names instead of crashing
This must go into the daemon (because that's the only entity that know that available interface names) and we must post an event / update the NodeState if an invalid interfaces name is specified

msherif1234:
before attaching XDP to interfaces we need to make
sure interfaces are valid and in the right state
fixed #91 

### DIFF
--- a/pkg/utils/interfaces_test.go
+++ b/pkg/utils/interfaces_test.go
@@ -1,0 +1,61 @@
+package utils
+
+import (
+	"net"
+	"testing"
+)
+
+func TestIsValidInterfaceNameAndState(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/utils/interfaces_test.go
```diff
@@ -0,0 +1,36 @@
+package utils
```
andreaskaris, line 1: For the test, it might be nice to make it mockable and not depend on actual values from the system, something like:
https://stackoverflow.com/questions/68940230/how-to-mock-net-interface

PR #104, opened 2022-08-11
API to list all interfaces with XDP prog attached

### LINKED ISSUE #98: Add API method to list all attached XDP interfaces
Add API method to list all attached XDP interfaces; it's possible that a daemon goes down, the resource is updated, and *then* the daemon comes back up but it has no way of knowing which interfaces are attached. When a daemon comes up it must compare the current resource state to the currently attached interfaces.

We must cover the following corner case:

i) process up, attaches watch
===
spec:
  interfaces: A,B
===
applied on A,B

ii) process down, no watch

iii) spec updated:
~~~
spec:
  interfaces: A
~~~

iv) processs up, attaches watch
applied on A

===================================
Compare that to what happens when all goes well:

i) process up, attaches watch
===
spec:
  interfaces: A,B
===
applied on A,B

ii) spec updated:
~~~
spec:
  interfaces: A
~~~

iii) watch sees change in spec, process gets old and new

iv) processs up
applied on A   (and b can be deleted)

msherif1234:
Fix #98 

### DIFF
--- a/bundle/manifests/ingressnodefirewall.openshift.io_ingressnodefirewallnodestates.yaml
+++ b/bundle/manifests/ingressnodefirewall.openshift.io_ingressnodefirewallnodestates.yaml
@@ -36,174 +36,180 @@ spec:
             description: IngressNodeFirewallNodeStateSpec defines the desired state
               of IngressNodeFirewallNodeState.
             properties:
-              ingress:
-                description: ingress is a list of ingress firewall policy rules.
-                items:
-                  description: IngressNodeFirewallRules define ingress node firewall
-                    rule.
-                  properties:
-                    rules:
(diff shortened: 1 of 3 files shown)


msherif1234:
unit-test returning `load program: operation not permitted (MEMLOCK may be too low, consider rlimit.RemoveMemlock)` I see the same when I ran ebpf ut I started a thread on cillium slack to findout more

msherif1234:
> unit-test returning `load program: operation not permitted (MEMLOCK may be too low, consider rlimit.RemoveMemlock)` I see the same when I ran ebpf ut I started a thread on cillium slack to findout more

need to use sudo -E not great I will drop unit-test :(

PR #109, opened 2022-08-12
Make ebpfSyncer and Firewall loader idempotent for add/update/delete

### LINKED ISSUE #101: ebpf rules -  implement correct idempotent rules merge
ebpf rules -  implement correct idempotent rules merge for node deletion, add and update

andreaskaris:
Fix #101 

**- How to verify it**

### DIFF
--- a/pkg/ebpf/ingress_node_firewall_loader.go
+++ b/pkg/ebpf/ingress_node_firewall_loader.go
@@ -2,27 +2,34 @@ package nodefwloader
 
 import (
 	"fmt"
+	"io/ioutil"
 	"log"
 	"net"
 	"os"
 	"path"
+	"reflect"
+	"regexp"
 	"strings"
 	"syscall"
 
+	"github.com/openshift/ingress-node-firewall/api/v1alpha1"
 	ingressnodefwiov1alpha1 "github.com/openshift/ingress-node-firewall/api/v1alpha1"
+	"github.com/openshift/ingress-node-firewall/pkg/interfaces"
 	"github.com/openshift/ingress-node-firewall/pkg/utils"
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/link"
 	"github.com/cilium/ebpf/rlimit"
+	apierrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog"
 )
 
 const (
-	xdpDeny     = 1 // XDP_DROP value
-	xdpAllow    = 2 // XDP_PASS value
-	bpfFSPath   = "/sys/fs/bpf"
-	xdpEBUSYErr = "device or resource busy"
+	xdpDeny                       = 1 // XDP_DROP value
+	xdpAllow                      = 2 // XDP_PASS value
+	bpfFSPath                     = "/sys/fs/bpf"
+	xdpIngressNodeFirewallProcess = "xdp_ingress_node_firewall_process"
+	linkSuffix                    = "_link"
 )
 
 // IngNodeFwController structure is the object hold controls for starting
@@ -39,14 +46,14 @@ type IngNodeFwController struct {
 // $BPF_CLANG and $BPF_CFLAGS are set by the Makefile.
 //go:generate bpf2go -cc $BPF_CLANG -cflags $BPF_CFLAGS -type ruleType_st -type event_hdr_st -type ruleStatistics_st Bpf ../../bpf/ingress_node_firewall_kernel.c -- -I ../../bpf/headers -I/usr/include/x86_64-linux-gnu/
 
-// NewIngNodeFwController creates new IngressNodeFirewall controller object
+// NewIngNodeFwController creates new IngressNodeFirewall controller object.
 func NewIngNodeFwController() (*IngNodeFwController, error) {
 	// Allow the current process to lock memory for eBPF resources.
 	if err := rlimit.RemoveMemlock(); err != nil {
 		return nil, err
 	}
 
-	pinDir := path.Join(bpfFSPath, "xdp_ingress_node_firewall_process")
+	pinDir := path.Join(bpfFSPath, xdpIngressNodeFirewallProcess)
 	if err := os.MkdirAll(pinDir, os.ModePerm); err != nil {
 		return nil, fmt.Errorf("failed to create pinDir %s: %s", pinDir, err)
 	}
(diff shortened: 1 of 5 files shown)


msherif1234:
lets agree or disagree why we try to purge/delete/update rules at the loader layer when I added this pkg I wanted to take rules either add or delete , interfaces attach/detach
doing merge at this layer doesn't sound right it has to be done at the controller level
unless u see advantage in doing here that I don't see ?

msherif1234:
@andreaskaris r handling add/update/delete at this layer because the maps are pinned so even when the process exists the maps persist ?

msherif1234:
pls include ut and e2e results

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -55,43 +59,295 @@ func NewIngNodeFwController() (*IngNodeFwController, error) {
...
+	if err != nil {
+		return fmt.Errorf("cannot get map info: %v", err)
+	}
+	klog.Infof("Ingress node firewall map Info: %+v with FD %s", info, infc.objs.BpfMaps.IngressNodeFirewallTableMap.String())
+
+	// Convert IngressNodeFirewallRules into data that can be written to the BPF map.
+	// Build a map of valid ebpfKeys pointing to the ebpfRules that should be associated to them.
+	ebpfKeyToRules := make(map[BpfLpmIpKeySt]BpfRulesValSt)
+	for interfaceName, ingressRules := range ifaceIngressRules {
+		// Look up the network interface by name.
```
msherif1234, line 102: use the new method in interfaces `GetInterfaceIndex`

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -166,113 +422,146 @@ func (infc *IngNodeFwController) makeIngressFwRulesMap(ingFirewallConfig ingress
...
+
+	// Lookup all keys inside the map and find keys that are stale.
+	var staleKeys []BpfLpmIpKeySt
+	var key BpfLpmIpKeySt
+	var value BpfRulesValSt
+	iterator := objs.BpfMaps.IngressNodeFirewallTableMap.Iterate()
+	for iterator.Next(&key, &value) {
+		keyFound := false
+		for _, desiredKey := range desiredKeys {
+			if desiredKey.IngressIfindex == key.IngressIfindex &&
```
msherif1234, line 482: why don't use DeepEqual()

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -55,43 +59,295 @@ func NewIngNodeFwController() (*IngNodeFwController, error) {
...
 
-	if err := infc.makeIngressFwRulesMap(ingFireWallConfig, isDelete, ifId); err != nil {
-		return fmt.Errorf("Failed to create map firewall rules: %v on if %d", err, ifId)
+	// Add/update all keys. This is an idempotent action and non-existing keys are added whereas existing keys
+	// are updated.
+	if err := infc.addOrUpdateRules(ebpfKeyToRules); err != nil {
+		return err
 	}
 
+	// Generate ingress node few events.
```
msherif1234, line 146: can u move this to NewController we just need to start evens loop once else just put FIXME and I will move it in future PRs

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -55,43 +59,295 @@ func NewIngNodeFwController() (*IngNodeFwController, error) {
...
 
 	return nil
 }
 
+// addOrUpdateRules is a small method containing this limited set of functionality to facilitate unit testing of
+// this part of code.
+func (infc *IngNodeFwController) addOrUpdateRules(ebpfKeyToRules map[BpfLpmIpKeySt]BpfRulesValSt) error {
+	for ebpfKey, ebpfRules := range ebpfKeyToRules {
+		log.Printf("Adding or updating ingress firewall rules for key %v", ebpfKey)
+		if err := infc.objs.BpfMaps.IngressNodeFirewallTableMap.Update(ebpfKey, ebpfRules, ebpf.UpdateAny); err != nil {
```
msherif1234, line 157: this is the part u said it doesn't work on slack ? pls add FIXME here for me to investigate so I won't forget about it

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -55,43 +59,295 @@ func NewIngNodeFwController() (*IngNodeFwController, error) {
...
+func (infc *IngNodeFwController) IngressNodeFwAttach(ifacesName ...string) error {
 	objs := infc.objs
-	rules := BpfRulesValSt{}
+	for _, ifaceName := range ifacesName {
+		if _, ok := infc.links[ifaceName]; ok {
+			klog.Infof("Interface %s is already attached and managed, skipping", ifaceName)
+			return nil
+		}
+		// Look up the network interface by name.
+		iface, err := net.InterfaceByName(ifaceName)
```
msherif1234, line 182: use the interfaces method to get ifId and use with the Attach

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -55,43 +59,295 @@ func NewIngNodeFwController() (*IngNodeFwController, error) {
...
+			Interface: iface.Index,
+		})
+		if err != nil {
+			if strings.Contains(err.Error(), xdpEBUSYErr) {
+				return fmt.Errorf("Interface %s is already attached", ifaceName)
+			}
+			return fmt.Errorf("could not attach XDP program: %s", err)
+		}
+		// Pin the XDP program.
+		lPinDir := path.Join(infc.pinPath, ifaceName+"_link")
```
msherif1234, line 199: lets use const for "_link" a suggested in the following comments

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -55,43 +59,295 @@ func NewIngNodeFwController() (*IngNodeFwController, error) {
...
+}
+
+// IngressNodeFwDetach detaches the eBPF program from the list of interfaces and cleans up the interfaces.
+// Additionally, it unloads all firewall rules that are associated to the interfaces.
+func (infc *IngNodeFwController) IngressNodeFwDetach(interfaceNames ...string) error {
+	// Detach from interfaces.
+	for _, interfaceName := range interfaceNames {
+		log.Printf("Detaching IngressNode Firewall program from interface %q", interfaceName)
+		if err := infc.cleanup(interfaceName); err != nil {
+			return err
```
msherif1234, line 216: lets aggr errors and return the aggr error here

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -55,43 +59,295 @@ func NewIngNodeFwController() (*IngNodeFwController, error) {
...
+// i)   Look up the network interface by name.
+// ii)  Attach the program to the interface.
+// iii) Pin the XDP program.
+func (infc *IngNodeFwController) IngressNodeFwAttach(ifacesName ...string) error {
 	objs := infc.objs
-	rules := BpfRulesValSt{}
+	for _, ifaceName := range ifacesName {
+		if _, ok := infc.links[ifaceName]; ok {
+			klog.Infof("Interface %s is already attached and managed, skipping", ifaceName)
+			return nil
```
msherif1234, line 179: we need to continue here for the next interfaces

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -55,43 +59,295 @@ func NewIngNodeFwController() (*IngNodeFwController, error) {
...
+		if err != nil {
+			if strings.Contains(err.Error(), xdpEBUSYErr) {
+				return fmt.Errorf("Interface %s is already attached", ifaceName)
+			}
+			return fmt.Errorf("could not attach XDP program: %s", err)
+		}
+		// Pin the XDP program.
+		lPinDir := path.Join(infc.pinPath, ifaceName+"_link")
+		if err := l.Pin(lPinDir); err != nil {
+			return fmt.Errorf("failed to pin link to pinDir %s: %s", lPinDir, err)
```
msherif1234, line 201: we need to append error to aggr errors slice and continue to the next interface and same for the above errors ?

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -55,43 +59,295 @@ func NewIngNodeFwController() (*IngNodeFwController, error) {
...
+		}
+
+		// Attach the program.
+		l, err := link.AttachXDP(link.XDPOptions{
+			Program:   objs.IngressNodeFirewallProcess,
+			Interface: iface.Index,
+		})
+		if err != nil {
+			if strings.Contains(err.Error(), xdpEBUSYErr) {
+				return fmt.Errorf("Interface %s is already attached", ifaceName)
```
msherif1234, line 194: EBUSY isn't an error right ? unless we aren't suppose to get this condition now in that case we need to remove teh special check for EBUSY and just error if err != nil

pkg/interfaces/interfaces.go
```diff
@@ -46,3 +47,33 @@ func GetInterfacesWithXDPAttached() ([]string, error) {
...
+
+	links, err := netlink.LinkList()
+	if err != nil {
+		return err
+	}
+	for _, l := range links {
+		if l.Attrs().Xdp != nil && l.Attrs().Xdp.Attached {
+			if err := netlink.LinkSetXdpFd(l, -1); err != nil {
+				errors = append(errors, err)
+			}
```
msherif1234, line 72: do u also need to set Attached bool to false ?

pkg/interfaces/interfaces.go
```diff
@@ -46,3 +47,33 @@ func GetInterfacesWithXDPAttached() ([]string, error) {
...
+		}
+	}
+	if len(errors) > 0 {
+		return fmt.Errorf("detach from all interfaces encountered errors: %v", errors)
+	}
+	return nil
```
msherif1234, line 78: return arrg error here

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -166,113 +422,146 @@ func (infc *IngNodeFwController) makeIngressFwRulesMap(ingFirewallConfig ingress
...
-		return fmt.Errorf("failed to find Link object for interface %s", ifName)
+// getStaleInterfaceKeys returns the keys for all rules that belong to stale interfaces, meaning interfaces
+// that are not attached any more.
+func (infc *IngNodeFwController) getStaleInterfaceKeys() ([]BpfLpmIpKeySt, error) {
+	objs := infc.objs
+
+	// Looup all valid interfaces IDs.
+	var validInterfaceIDs []uint32
+	for interfaceName := range infc.links {
+		iface, err := net.InterfaceByName(interfaceName)
```
msherif1234, line 513: use the new interfaces method here as well

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -55,43 +59,295 @@ func NewIngNodeFwController() (*IngNodeFwController, error) {
...
+	// Delete all keys that should be deleted.
+	if err := infc.purgeKeys(staleKeys); err != nil {
+		klog.Infof("Purge keys operation encountered issues, err: %q", err)
+	}
+
+	return nil
+}
+
+// GetBPFMapContent lists all existing keys and rules inside the map. Used for unit testing.
+func (infc *IngNodeFwController) GetBPFMapContent() (map[BpfLpmIpKeySt]BpfRulesValSt, error) {
```
msherif1234, line 242: Can u add `Test` to the name so we are sure this only for unit-test

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -55,43 +59,295 @@ func NewIngNodeFwController() (*IngNodeFwController, error) {
...
+
+	return keysToRules, nil
+}
+
+// CleaneBPFObjs closes the current objs and removes all interface pins.
+func (infc *IngNodeFwController) CleaneBPFObjs() error {
+	if err := infc.objs.Close(); err != nil {
+		return fmt.Errorf("failed to close eBPF objs err: %q", err)
+	}
+	return nil
```
msherif1234, line 266: shouldn't we also RemoveAllPins() at this point ?

pkg/ebpfsyncer/ebpfsyncer.go
```diff
@@ -80,93 +84,176 @@ func (e *ebpfSingleton) SyncInterfaceIngressRules(
...
 		}
-		ifId := uint32(iface.Index)
+	}
+
+	e.log.Info("Running XDP detach operation from all interfaces")
+	if err := intutil.DetachXDPFromAllInterfaces(); err != nil {
+		e.log.Info("Error during detach operation", "err", err)
+	}
+
+	e.log.Info("Removing all pins")
```
msherif1234, line 180: why don't we combine all those in one method to export less details to this pkg ? it doesn't need to know anything about pinDir etc ?

pkg/ebpfsyncer/ebpfsyncer.go
```diff
@@ -80,93 +84,176 @@ func (e *ebpfSingleton) SyncInterfaceIngressRules(
...
-				return err
-			}
-		}
-		e.managedInterfaces = make(map[string]struct{})
-		err := e.c.CleaneBPFObjs()
-		if err != nil {
-			return err
-		}
-		e.c = nil
+		return e.resetAll()
```
msherif1234, line 95: isDelete can be set if we just deleting one rules obj but not all objs right ? so we can't resetAll()

pkg/ebpfsyncer/ebpfsyncer.go
```diff
@@ -80,93 +84,176 @@ func (e *ebpfSingleton) SyncInterfaceIngressRules(
...
+
+	// Load IngressNodeFirewall Rules (this is idempotent and will add new rules and purge rules that shouldn't exist).
+	if err := e.loadIngressNodeFirewallRules(ifaceIngressRules); err != nil {
+		return err
+	}
+	return nil
+}
+
+// getBPFMapContent lists the content of the current BPF map. Used for unit testing only.
+func (e *ebpfSingleton) getBPFMapContent() (map[nodefwloader.BpfLpmIpKeySt]nodefwloader.BpfRulesValSt, error) {
```
msherif1234, line 121: same here pls add Test to the name to make it clear its ut only

msherif1234 (commented):
lets get started with this I need to review the unit-test after

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -55,43 +59,295 @@ func NewIngNodeFwController() (*IngNodeFwController, error) {
...
+	infc := &IngNodeFwController{
 		objs:    objs,
 		pinPath: pinDir,
 		links:   make(map[string]link.Link, 0),
-	}, nil
+	}
+	// Load pinned links from /sys/fs/bpf/xdp_ingress_node_firewall_process on initialization.
+	// That way, the state in /sys/fs/bpf/xdp_ingress_node_firewall_process and the tracked list of links
+	// will be in sync.
+	if err := infc.loadPinnedLinks(); err != nil {
```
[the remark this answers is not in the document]
andreaskaris: So right now it's only called a single time over the life time of the process, unless all rules are deleted at which time its reset to nil:

a) -node-firewall/controllers/ingressnodefirewallnodestate_controller.go
~~~
4     if err := ebpfsyncer.GetEbpfSyncer(ctx, r.Log, r.Stats, mock).SyncInterfaceIngressRules(instance.Spec.Interf      aceIngressRules, isDelete); err != nil {
   95         return ctrl.Result{}, errors.Wrapf(err, "FailedToSyncIngressNodeFirewallResources")
   96     } 
~~~

b)   
~~~
 37 func GetEbpfSyncer(ctx context.Context, log logr.Logger, stats *metrics.Statistics, mock EbpfSyncer) EbpfSyncer       {       
   38     once.Do(func() {
   39         // Check if instace is nil. For mock tests, one can provide a custom instance.
   40         if mock == nil {
   41             instance = &ebpfSingleton{
   42                 ctx:               ctx,
   43                 log:               log,
   44                 stats:             stats,
   45                 managedInterfaces: make(map[string]struct{}),
   46             }
   47         } else {
   48             instance = mock
   49         }
   50     })
   51     return instance
   52 }
(...)
   87     // Create a new manager if none exists.
   88     if err := e.createNewManager(); err != nil { 
   89         return err
   90     }
(...)
  128 // Create a new manager if none exists.
  129 func (e *ebpfSingleton) createNewManager() error {
  130     var err error
  131     if e.c == nil {
  132         e.log.Info("Creating a new eBPF firewall node controller")
  133         if e.c, err = nodefwloader.NewIngNodeFwController(); err != nil {
  134             return fmt.Errorf("Failed to create nodefw controller instance, err: %q", err)
  135         }
  136     }
  137     return nil
  138 }
(...)
~~~

So any follow-up call here will not create a new object, only if it was reset on purpose.
And then that instance is only set to nil when all rules and objects are deleted for the given node.

loadPinnedLinks() goes through the list of pins in /sys/fs/bpf/xdp_ingress_node_firewall_process and re-creates in-memory Link object and readd thems to links.Links

pkg/ebpf/ingress_node_firewall_loader.go
andreaskaris: done

pkg/ebpf/ingress_node_firewall_loader.go
andreaskaris: Done.

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -166,113 +422,146 @@ func (infc *IngNodeFwController) makeIngressFwRulesMap(ingFirewallConfig ingress
...
+	for _, keyToDelete := range keys {
+		klog.Infof("Purging key %v", keyToDelete)
+		err := objs.BpfMaps.IngressNodeFirewallTableMap.Delete(keyToDelete)
+		if err != nil {
+			errors = append(errors, err)
+		}
 	}
-	if err := os.RemoveAll(infc.pinPath); err != nil {
-		return fmt.Errorf("failed to remove pinpath %s err: %q", infc.pinPath, err)
+	if len(errors) > 0 {
```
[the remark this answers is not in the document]
andreaskaris: Oh that's cool, I didn't know about that one!

pkg/ebpf/ingress_node_firewall_loader.go
andreaskaris: I moved it

pkg/ebpf/ingress_node_firewall_loader.go
andreaskaris: done

pkg/ebpf/ingress_node_firewall_loader.go
andreaskaris: done

pkg/ebpf/ingress_node_firewall_loader.go
andreaskaris: done

pkg/ebpf/ingress_node_firewall_loader.go
andreaskaris: *facepalm* Good catch!

Makes sense that I missed this as this is only called with  a single interface atm:
~~~
IngressNodeFwAttach(intf)
~~~

pkg/ebpf/ingress_node_firewall_loader.go
andreaskaris: done

pkg/ebpf/ingress_node_firewall_loader.go
andreaskaris: I removed that check as with the new "scan the folder and create link objects from pins", the dual attach doesn't happen any more.

On the other hand, I used to run into race conditions between detach/attach operations during the unit tests so I'm trying to catch those here:
~~~
  204 func (e *ebpfSingleton) attachNewInterfaces(ifaceIngressRules map[string][]v1alpha1.IngressNodeFirewallRules) error {   
  205     for intf := range ifaceIngressRules {                                                                               
  206         // First, check if the interface name is valid.                                                                 
  207         if !isValidInterfaceNameAndState(intf) {                                                                        
  208             return fmt.Errorf("Fail to attach ingress node Firewall rules: invalid interface %s", intf)                 
  209         }                                                                                                               
  210                                                                                                                         
  211         // Then, check if the interface is already managed.                                                             
  212         if _, ok := e.managedInterfaces[intf]; !ok {                                                                    
  213             // Attach to the interfaces - in case the interface is already attached, retry.                             
  214             err := retry.OnError(                                                                                       
  215                 retry.DefaultRetry,                                                                                     
  216                 func(err error) bool {                                                                                  
  217                     return strings.Contains(err.Error(), "device or resource busy")                                     
  218                 },                                                                                                      
  219                 func() error {                                                                                          
  220                     e.log.Info("Attaching firewall interface", "intf", intf)                                            
  221                     if err := e.c.IngressNodeFwAttach(intf); err != nil {                                               
  222                         e.log.Error(err, "Fail to attach ingress firewall prog")                                        
  223                         return err                                                                                      
  224                     }                                                                                                   
  225                     e.managedInterfaces[intf] = struct{}{}                                                              
  226                     return nil                                                                                          
  227                 })                                                                                                      
  228             if err != nil {                                                                                             
  229                 return err                                                                                              
  230             }                                                                                                           
  231         }                                                                                                               
  232     }                                                                                                                   
  233     return nil                                                                                                          
  234 }       
~~~

But I'm having a hard time replicating that issue at the moment.

pkg/interfaces/interfaces.go
andreaskaris: I'm not even sure that we really need that function at all.

The only place where I'm using it is here:
~~~
  165 // resetAll deletes all current attachments and cleans all eBPFObjects. It then sets the ingress firewall manager       
  166 // back to nil. It also deletes all pins and removed all XDP attachments for all system interfaces.                     
  167 // FIXME: During tests, I encountered issues with detach operations which did not work in all cases. This resetAll()    
  168 // here runs some redundant steps such as DetachXDPFromAllInterfaces which might not be necessary but until we          
  169 // get to the bottom of the detach issues, I opted for doing more than needed.                                          
  170 func (e *ebpfSingleton) resetAll() error {                                                                              
  171     e.log.Info("Running detach operation of managed interfaces")                                                        
  172     for intf := range e.managedInterfaces {                                                                             
  173         err := e.c.IngressNodeFwDetach(intf)                                                                            
  174         if err != nil {                                                                                                 
  175             e.log.Info("Could not detach managed interface", "intf", intf, "err", err)                                  
  176         }                                                                                                               
  177     }                                                                                                                   
  178                                                                                                                         
  179     e.log.Info("Removing all pins")                                                                                     
  180     if err := nodefwloader.RemoveAllPins(); err != nil {                                                                
  181         e.log.Info("Could not remove all eBPF pins", "err", err)                                                        
  182     }                                                                                                                   
  183                                                                                                                         
  184     e.log.Info("Removing table map")                                                                                    
  185     if err := nodefwloader.RemoveTableMap(); err != nil {                                                               
  186         e.log.Info("Could not remove eBPF table map", "err", err)                                                       
  187     }                                                                                                                   
  188                                                                                                                         
  189     e.log.Info("Running cleanup of eBPF objects")                                                                       
  190     if err := e.c.CleaneBPFObjs(); err != nil {                                                                         
  191         e.log.Info("Could not clean eBPF objects", "err", err)                                                          
  192     }                                                                                                                   
  193                                                                                                                         
  194     // FIXME: Is this really necesary? The steps before should have cleaned up everything already.                      
  195     e.log.Info("Running XDP detach operation from all interfaces")                                                      
  196     if err := intutil.DetachXDPFromAllInterfaces(); err != nil {                                                        
  197         e.log.Info("Error during detach operation", "err", err)                                                         
  198     }                                                                                                                   
  199                                                                                                                         
  200     e.managedInterfaces = make(map[string]struct{})                                                                     
  201     e.c = nil                                                                                                           
  202                                                                                                                         
  203     return nil                                                                                                          
  204 }   
~~~

The resetAll method above isn't really ideal either, it's a result of my confusion about why the links do not detach correctly.

pkg/interfaces/interfaces.go
andreaskaris: done

pkg/ebpf/ingress_node_firewall_loader.go
andreaskaris: done

pkg/ebpf/ingress_node_firewall_loader.go
andreaskaris: done

pkg/ebpf/ingress_node_firewall_loader.go
andreaskaris: I broke the original method into 3 individual chunks to make testing and moving around the code easier (also in the future). That was part of me trying to fix the detach stuff:
~~~
  179     e.log.Info("Removing all pins")
  180     if err := nodefwloader.RemoveAllPins(); err != nil {
  181         e.log.Info("Could not remove all eBPF pins", "err", err)
  182     }
  183 
  184     e.log.Info("Removing table map")
  185     if err := nodefwloader.RemoveTableMap(); err != nil {
  186         e.log.Info("Could not remove eBPF table map", "err", err)
  187     }
  188 
  189     e.log.Info("Running cleanup of eBPF objects")
  190     if err := e.c.CleaneBPFObjs(); err != nil {
  191         e.log.Info("Could not clean eBPF objects", "err", err)
  192     }
~~~

Is that here ok for you? Otherwise, I'll make them non-public and move them back into the same public method  CleaneBPFObjs:
~~~
  277 // CleaneBPFObjs closes the current objs and removes all interface pins.
  278 // FIXME: Once we figure out why links don't close / detach correctly, possibly
  279 // recombine CleaneBPFObjs / RemoveAllPins / RemoveTableMap into a single method.
  280 func (infc *IngNodeFwController) CleaneBPFObjs() error { 
  281     if err := infc.objs.Close(); err != nil { 
  282         return fmt.Errorf("failed to close eBPF objs err: %q", err) 
  283     } 
  284     return nil
  285 } 
  286 
  287 // RemoveAllPins removes all pins for XDP.
  288 func RemoveAllPins() error { 
  289     pinDir := path.Join(bpfFSPath, xdpIngressNodeFirewallProcess) 
  290     files, err := ioutil.ReadDir(pinDir) 
  291     if err != nil { 
  292         return err
  293     } 
  294 
  295     for _, file := range files { 
  296         matched, err := regexp.Match(".*"+linkSuffix+"$", []byte(file.Name())) 
  297         if err != nil { 
  298             return err
  299         } 
  300         if matched { 
  301             if err := os.Remove(path.Join(pinDir, file.Name())); err != nil { 
  302                 return err
  303             } 
  304         } 
  305     } 
  306     return nil
  307 } 
  308 
  309 // RemoveTableMap removes the ebpf table map.
  310 func RemoveTableMap() error { 
  311     pinDir := path.Join(bpfFSPath, xdpIngressNodeFirewallProcess) 
  312     return os.Remove(path.Join(pinDir, "ingress_node_firewall_table_map")) 
  313 } 
~~~

pkg/ebpfsyncer/ebpfsyncer.go
andreaskaris: See my comment aboeve but agreed, I'll move them back into one method but I'll keep them modular as private methods for the future just in case we need to troubleshoot this further

andreaskaris: done

pkg/ebpf/ingress_node_firewall_loader.go
andreaskaris: done

pkg/ebpfsyncer/ebpfsyncer.go
andreaskaris: With this PR, `isDelete` is only called when the `IngressNodeFirewallNodeState` object is deleted for that node, so this is a global delete operation when the reconciler reports that the object gets a delete.

For object creation or updates:
All rules are applied in an idempotent way, meaning that we go through all rules and diff what should be there vs what's there and deletion of individual rules is achieved that way.

andreaskaris:
~~~
   80 // IngressNodeFwRulesLoader adds/updates/deletes ingress node firewall rules to the eBPF LPM MAP in an idempoten      t way.  
   81 // IngressNodeFwRulesLoader executes the following actions in order:
   82 // i)   Get eBPF objs to create/update eBPF maps and get map info.
   83 // ii)  Build a map of valid ebpfKeys pointing to the ebpfRules that should be associated to them (built from
   84 //      ifaceIngressRules).
   85 // iii) Get stale keys (= keys inside the eBPF map but not inside the currently desired ruleset).
   86 // iv)  Purge all stale keys from the eBPF map.
   87 // v)   Add/update all keys. This is an idempotent action and non-existing keys are added whereas existing keys
   88 //      are updated.
   89 // vi)  Generate ingress node firewall events.
   90 // In the context of this method, stale keys are keys that figure inside the eBPF map but that are not generated
   91 // during step ii) from the provided ingressRules slice.
   92 func (infc *IngNodeFwController) IngressNodeFwRulesLoader(
   93     ifaceIngressRules map[string][]v1alpha1.IngressNodeFirewallRules) error {
~~~

pkg/ebpfsyncer/ebpfsyncer.go
andreaskaris: done

pkg/interfaces/interfaces.go
msherif1234: can we remove unneeded function then

msherif1234:
/LGTM nice work on the loader and the loader ut Thanks for all your great efforts!!

andreaskaris:
I squashed the 2 commits together :)

PR #115, opened 2022-08-16
create-kind-cluster now deploys ingress controller as well

andreaskaris:
Signed-off-by: Andreas Karis <ak.karis@gmail.com>

**- How to verify it**


hack/kind-cluster.sh
```diff
@@ -32,5 +34,28 @@ nodes:
...
+  make docker-build-daemon DAEMON_IMG=localhost/ingress-node-firewall-daemon:local
+  kind load docker-image localhost/ingress-node-firewall-daemon:local
+  if ! command -v yq ; then 
+      echo "Cannot find command yq."
+      echo "Please modify file config/manager/env.yaml manually and point DAEMONSET_IMAGE to:"
+      echo "    localhost/ingress-node-firewall-daemon:local"
+      echo "Then, hit any key"
+      read
+  fi
+  yq e '.spec.template.spec.containers[0].env[0].value = "localhost/ingress-node-firewall-daemon:local"'  -i config/manager/env.yaml
```
msherif1234, line 53: hard coding container idx and env idx to match env.yaml will cause trouble if we reorder anything in env.yaml by accident can we find the index by lookup the env name ?

andreaskaris:
@msherif1234  @martinkennelly  lmk wdyt

hack/kind-cluster.sh
```diff
@@ -34,6 +42,58 @@ nodes:
...
+# install_operator installs the operator on top of the kind cluster
+install_operator() {
+  pushd ${DIR}/../
+  make install
+  make docker-build IMG=localhost/ingress-node-firewall-controller:local
+  kind load docker-image localhost/ingress-node-firewall-controller:local
+  make docker-build-daemon DAEMON_IMG=localhost/ingress-node-firewall-daemon:local
+  kind load docker-image localhost/ingress-node-firewall-daemon:local
+  set +x
+  hack/set-daemon-image.sh localhost/ingress-node-firewall-daemon:local || then_hit_any_key
```
msherif1234, line 56: instead of this we can use yq and replace the image directly ?

msherif1234:
I will also leave the manual steps in the readme and leave it up to the user to use the new rule or use the manual steps

hack/kind-cluster.sh
andreaskaris: that is using yq under the hood though (plus it installs yq, etc.)

andreaskaris: The content of that script is:
~~~
[akaris@linux ingress-node-firewall (akarislibxdp)]$ cat hack/set-daemon-image.sh
#!/bin/bash

set -eux

DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
DAEMONSET_IMAGE=$1

if ! command -v yq ; then 
  echo "Could not find yq binary - trying to install it"
  go install github.com/mikefarah/yq/v4@latest
fi

if ! command -v yq ; then 
  echo "Cannot find command yq and could not install it."
  echo "Please modify file config/manager/env.yaml manually and point DAEMONSET_IMAGE to:"
  echo "    ${DAEMONSET_IMAGE}"
fi
yq e "(.spec.template.spec.containers[] | select(.name == \"manager\") .env[] | select(.name == \"DAEMONSET_IMAGE\") ).value=\"${DAEMONSET_IMAGE}\"" \
    -i ${DIR}/../config/manager/env.yaml
~~~

Keeping the script as an individual tool is IMO better than merging it into the kind-cluster.sh script, but that's my personal opinion

andreaskaris:
> I will also leave the manual steps in the readme and leave it up to the user to use the new rule or use the manual steps

Sounds good I'll address that upon return to the office

hack/kind-cluster.sh
andreaskaris: I have a relatively strong preference for keeping the yq stuff inside set-daemon-image.sh ; especially given that we are keeping the manual steps in our documentation. If your indifferent about it, I'd like to keep it that way. If you have a strong preference for integrating that in kind-cluster.sh, then I'll add the logic to kind-cluster.sh as well (at the cost of having the same logic both in set-daemon-image.sh and kind-cluster.sh). lemme know what you think :-)

msherif1234: i was referring to the part where the script pause waiting to enter something

PR #116, opened 2022-08-16
Revert "makefile fixes"

msherif1234:
Reverts openshift/ingress-node-firewall#112


msherif1234:
make test failed with this changes I will revert them for now

PR #120, opened 2022-08-18
 E2E: Remove hardcode KinD's CIDR range from e2e fw rules

msherif1234:
we need to be able to run e2e test on any cluster not KinD cluster only

**- How to verify it**
- run `make test-e2e` on any cluster CIDR range should be calculated based on the nodes IPv4and IPv6 addresses.
- manual run of e2e passed on kind
```

Ran 10 of 10 Specs in 304.583 seconds
SUCCESS! -- 10 Passed | 0 Failed | 0 Pending | 0 Skipped
```

drive CIDR value for the rules from node IPv4 and IPv6 addresses

### DIFF
--- a/test/e2e/ingress-node-firewall/ingress-node-firewall.go
+++ b/test/e2e/ingress-node-firewall/ingress-node-firewall.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net"
 	"os"
 	"os/exec"
 	"time"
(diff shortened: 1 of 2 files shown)


test/e2e/ingress-node-firewall/ingress-node-firewall.go
```diff
@@ -105,6 +106,29 @@ func NodesIPs(nodes []v1.Node) []string {
...
+		if addr.To4() != nil {
+			v4Mask := net.CIDRMask(v4CIDRLen, 32)
+			v4CIDR = fmt.Sprintf("%s/%d", addr.Mask(v4Mask), v4CIDRLen)
+		} else if addr.To16() != nil {
+			v6Mask := net.CIDRMask(v6CIDRLen, 128)
+			v6CIDR = fmt.Sprintf("%s/%d", addr.Mask(v6Mask), v6CIDRLen)
+		} else {
+			return "", "", fmt.Errorf("invalid ip address family %s", ip)
+		}
+		if v4CIDR != "" && v6CIDR != "" {
```
martinkennelly, line 125: how can you expect to get a v6CIDR if its v4 only cluster?

msherif1234: We will use dual stack cluster for our ci e2e run

PR #121, opened 2022-08-18
E2E: Add platform pkg to detect running on OCP cluster type

msherif1234:

add pkg to detect if running on OCP cluster or not to customize securityContext

**- How to verify it**
e2e was extended to check the cluster type

### DIFF
--- a/config/openshift/namespace.yaml
+++ b/config/openshift/namespace.yaml
@@ -10,5 +10,5 @@ metadata:
     openshift.io/node-selector: ""
     openshift.io/description: "Openshift ingress node firewall components"
     workload.openshift.io/allowed: "management"
-  name: system
+  name: openshift-ingress-node-firewall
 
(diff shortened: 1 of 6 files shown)


config/openshift/namespace.yaml
```diff
@@ -10,5 +10,5 @@ metadata:
     openshift.io/node-selector: ""
     openshift.io/description: "Openshift ingress node firewall components"
     workload.openshift.io/allowed: "management"
-  name: system
+  name: openshift-ingress-node-firewall
```
martinkennelly, line 13: shouldnt this be replaced by kustomize anyway? You found it wasn't being replaced?

martinkennelly:
LGTM - curious, did you come up with the idea of looking for routes API or is that a standard pattern to check if we are using OCP?

config/openshift/namespace.yaml
msherif1234: Ns wasn’t created when ran on ocp I thought kustomize will use this yaml to create namespace if my understanding is incorrect let me know

msherif1234:
> LGTM - curious, did you come up with the idea of looking for routes API or is that a standard pattern to check if we are using OCP?

Yeah standard pattern used by many operators

PR #123, opened 2022-08-19
SDN-3328 - Log dropped firewall events to a syslog-based sidecar

frobware:
Log the daemon's firewall events via syslog. The daemonset now deploys an additional syslog container named "events". 

The purpose of this change is two fold:

1) Log dropped events to `stdout` (the new syslog "events" container logs to `stdout`) so that the kubelet will take care of log rotation. Currently, dropped events are logged to a file in `/tmp` in the daemon container.

2) Separate dropped events from "operational" daemon events. The "events" container is now just a stream of dropped events; the "daemon" container carries all operational/reconcile-style messages which makes it easier to debug and post-process the set of all dropped events.

I chose to create a very simple syslog server, based on https://github.com/mcuadros/go-syslog. I think this approach simplifies the deployment (for both KIND and OpenShift) as you don't have to install any rsyslog package. And, as we own both pieces, (the syslog binary, the daemon binary) we don't have to create any configmap for `rsyslog`; the daemon and the syslog containers rendezvous through the named socket `/var/run/syslog`. 

Note: logging using Go's https://pkg.go.dev/log/syslog package forces us to use one of "/dev/log", "/var/run/syslog", or "/var/run/log" and that is why I chose "/var/run/syslog".

The format of events is:

```console
$ oc logs -f -c events -n ingress-node-firewall-system ingress-node-firewall-daemon-lnrf8
2022-08-19 08:00:07 +0000 UTC kind-worker2 daemon ruleId 10 action Drop len 98 if eth0
2022-08-19 08:00:07 +0000 UTC kind-worker2 daemon 	ipv4 src addr 172.18.0.1 dst addr 172.18.0.4
2022-08-19 08:00:07 +0000 UTC kind-worker2 daemon 	icmpv4 type 8 code 0
2022-08-19 08:00:17 +0000 UTC kind-worker2 daemon ruleId 10 action Drop len 118 if eth0
2022-08-19 08:00:17 +0000 UTC kind-worker2 daemon 	ipv6 src addr fc00:f853:ccd:e793::1 dst addr fc00:f853:ccd:e793::4
2022-08-19 08:00:17 +0000 UTC kind-worker2 daemon 	icmpv6 type 128 code 0
```

PR #135, opened 2022-08-22
Add "events" container to firewall config unit tests

frobware:
This is a follow-up to PR #123.

### DIFF
--- a/controllers/ingressnodefirewallconfig_controller_test.go
+++ b/controllers/ingressnodefirewallconfig_controller_test.go
@@ -25,8 +25,12 @@ var _ = Describe("Ingress nodefirewall config Controller", func() {
 					Namespace: IngressNodeFwConfigTestNameSpace,
 				},
 			}
+			daemonInitContainers := map[string]string{
+				"mount-bpffs": "test-daemon:latest",
+			}
 			daemonContainers := map[string]string{
 				"daemon":          "test-daemon:latest",
+				"events":          "test-daemon:latest",
 				"kube-rbac-proxy": "kube-rbac-proxy:latest",
 			}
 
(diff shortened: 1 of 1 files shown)


frobware:
> Nice while you are here let’s add the initContainer too to ut

Done in https://github.com/openshift/ingress-node-firewall/compare/f7defac63446db3c3fb13528d7cac10110c8c9df..c3372714c7382673aaa25572102f07151c8341f3.

controllers/ingressnodefirewallconfig_controller_test.go
```diff
@@ -26,7 +26,9 @@ var _ = Describe("Ingress nodefirewall config Controller", func() {
 				},
 			}
 			daemonContainers := map[string]string{
+				"mount-bpffs":     "test-daemon:latest",
```
msherif1234, line 29: this won't work we need to create new map for daemonInitContainers and add the mount-bpffs there 
then new loop to go over `daemonSet.Spec.Template.Spec.InitContainers`
something like this
```golang
for _, c := range daemonSet.Spec.Template.Spec.InitContainers {
    image, ok := daemonInitContainers[c.Name]
    Expect(ok).To(BeTrue(), fmt.Sprintf("init container %s not found in %s", c.Name, daemonInitContainers))
    Expect(c.Image).To(Equal(image))
}
```

frobware: OK. I didn't look to closely when I added the initContainer name. Thanks.

frobware: Done in https://github.com/openshift/ingress-node-firewall/compare/c3372714c7382673aaa25572102f07151c8341f3..46f5a27d3c523a4ad2439511ddeaaefcb6dfcb92.

controllers/ingressnodefirewallconfig_controller_test.go
```diff
@@ -41,10 +45,16 @@ var _ = Describe("Ingress nodefirewall config Controller", func() {
...
 			}, 2*time.Second, 200*time.Millisecond).ShouldNot((HaveOccurred()))
 			Expect(daemonSet).NotTo(BeZero())
-			Expect(len(daemonSet.Spec.Template.Spec.Containers)).To(BeNumerically(">", 0))
-			for _, c := range daemonSet.Spec.Template.Spec.Containers {
-				image, ok := daemonContainers[c.Name]
-				Expect(ok).To(BeTrue(), fmt.Sprintf("container %s not found in %s", c.Name, daemonContainers))
+			Expect(daemonSet.Spec.Template.Spec.InitContainers).To(HaveLen(len(daemonInitContainers)))
+			Expect(daemonSet.Spec.Template.Spec.Containers).To(HaveLen(len(daemonContainers)))
+			for name, image := range daemonInitContainers {
+				ok, c := findContainerByName(name, daemonSet.Spec.Template.Spec.InitContainers)
```
msherif1234, line 51: you don't really need another loop here daemonCOntainers is hash map u provide the name and u get the result ?

frobware: Not sure I understand; I want iterate over the expected containers and verify that each named container exists in `Spec.InitContainers`, which is a slice.

frobware: > you don't really need another loop here daemonCOntainers is hash map u provide the name and u get the result ?

Putting it another way: we should iterate over the expectations. The way the test was previously allowed me to add the init container reference "mount-bpffs" to `daemonContainers`, yet the test didn't fail. And I got caught out by that because it should have failed.

msherif1234: `findContainerByName` is doing loop for no reason u can lookup the map by name and get ur answer ?

frobware: OK - as we have the new len assertion we can iterate as it was previously.

frobware: Done in https://github.com/openshift/ingress-node-firewall/compare/46f5a27d3c523a4ad2439511ddeaaefcb6dfcb92..ec4fb75a0efe555f259c49584a1a10633ce82b91.

PR #140, opened 2022-08-23
Fix OCP deploy script

sabinaaledort:


openshift-ci/deploy_ingress_node_firewall.sh
```diff
@@ -106,6 +109,8 @@ fi
 
 ./wait_for_csv.sh
 
+oc label ns openshift-marketplace --overwrite pod-security.kubernetes.io/enforce=baseline
```
msherif1234, line 112: so why we needed this now ?

sabinaaledort: to fix this error:
```
+(./deploy_ingress_node_firewall.sh:58): main(): oc apply -f -
Error from server (Forbidden): error when creating "STDIN": pods "buildindex" is forbidden: violates PodSecurity "restricted:v1.24": privileged (container "priv" must not set securityContext.privileged=true), allowPrivilegeEscalation != false (container "priv" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "priv" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "priv" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "priv" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

frobware:
I think the script should set some best practice flags:

```sh
set -eu
set -o pipefail
```

msherif1234:
> 

we can do that once this merged w/o this PR the test run for very long time before it fails

frobware:
> > 
> 
> we can do that once this merged w/o this PR the test run for very long time before it fails

Ack. (I was late to this.)

I would also advocate using shellcheck and fix most/all/some of the warnings. 

For example, the following was flagged:

```sh
In openshift-ci/deploy_ingress_node_firewall.sh line 137:
  if [[ ${desired_ds_num} > 1 ]] && [[ ${ready_ds_num} == ${desired_ds_num} ]]; then
                          ^-- SC2071 (error): > is for string comparisons. Use -gt instead.
                                                          ^---------------^ SC2053 (warning): Quote the right-hand side of == in [[ ]] to prevent glob matching.
```

sabinaaledort:
> > > 
> > 
> > 
> > we can do that once this merged w/o this PR the test run for very long time before it fails
> 
> Ack. (I was late to this.)
> 
> I would also advocate using shellcheck and fix most/all/some of the warnings.
> 
> For example, the following was flagged:
> 
> ```shell
> In openshift-ci/deploy_ingress_node_firewall.sh line 137:
>   if [[ ${desired_ds_num} > 1 ]] && [[ ${ready_ds_num} == ${desired_ds_num} ]]; then
>                           ^-- SC2071 (error): > is for string comparisons. Use -gt instead.
>                                                           ^---------------^ SC2053 (warning): Quote the right-hand side of == in [[ ]] to prevent glob matching.
> ```

thanks for introducing me the shellcheck tool! i fixed all the warnings.

sabinaaledort:
> I think the script should set some best practice flags:
> 
> ```shell
> set -eu
> set -o pipefail
> ```

thanks, added.

msherif1234:
ci will fail because of known QEMU issue

PR #142, opened 2022-08-23
Dockerfile: use vendor directory when building

frobware:
Avoid the call to 'go mod download' as that doesn't necessarily give
us a reproducible build. In Dockerfile.daemon we already copy the
vendor directory so we should be consistent.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,9 +5,8 @@ WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY go.mod go.mod
 COPY go.sum go.sum
-# cache deps before building and copying source so that we don't need to re-download as much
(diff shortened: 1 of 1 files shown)


frobware:
/hold
Need to test this (tomorrow).

msherif1234:
FYI this dockerfile for the operator doesn't use vendor while dockerfile.Openshift does, and this Dockerfile only for KinD testing

frobware:
> FYI this dockerfile for the operator doesn't use vendor while dockerfile.Openshift does, and this Dockerfile only for KinD testing

I can close this. I was just trying to ensure that the bits I'm building are largely the same. (And anything that involves downloading from the internet is very slow given my bandwidth.)

PR #148, opened 2022-08-24
Trigger nodeState reconcilation only for fwobj with matching node's label

msherif1234:

- trigger nodeState reconciler only if fwobj nodeselector match node object label
- add `make test-race` to detect race conditions.

**- How to verify it**
- ran UT
- ran e2e on kind and it passed with those changes
```

Ran 11 of 11 Specs in 331.492 seconds
SUCCESS! -- 11 Passed | 0 Failed | 0 Pending | 0 Skipped
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,6 +127,10 @@ vet: ## Run go vet against code.
 test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(ENVTEST_ASSETS_DIR)/bin" go test ./... -coverprofile cover.out
 
+.PHONY: test-race
+test-race: manifests generate fmt vet envtest ## Run tests and check for race conditions.
+	KUBEBUILDER_ASSETS="$(ENVTEST_ASSETS_DIR)/bin" go test -race ./...
+
 .PHONY: create-kind-cluster 
 create-kind-cluster: ## Create a kind cluster.
 	hack/kind-cluster.sh
--- a/config/crd/bases/ingressnodefirewall.openshift.io_ingressnodefirewallconfigs.yaml
+++ b/config/crd/bases/ingressnodefirewall.openshift.io_ingressnodefirewallconfigs.yaml
@@ -96,8 +96,8 @@ spec:
                   description: "Condition contains details for one aspect of the current
                     state of this API Resource. --- This struct is intended for direct
                     use as an array at the field path .status.conditions.  For example,
-                    type FooStatus struct{ // Represents the observations of a foo's
-                    current state. // Known .status.conditions.type are: \"Available\",
+                    \n type FooStatus struct{ // Represents the observations of a
+                    foo's current state. // Known .status.conditions.type are: \"Available\",
                     \"Progressing\", and \"Degraded\" // +patchMergeKey=type // +patchStrategy=merge
                     // +listType=map // +listMapKey=type Conditions []metav1.Condition
                     `json:\"conditions,omitempty\" patchStrategy:\"merge\" patchMergeKey:\"type\"
--- a/controllers/ingressnodefirewall_controller.go
+++ b/controllers/ingressnodefirewall_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	"github.com/go-logr/logr"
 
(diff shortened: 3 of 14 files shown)


config/crd/bases/ingressnodefirewall.openshift.io_ingressnodefirewallconfigs.yaml
```diff
@@ -96,8 +96,8 @@ spec:
                   description: "Condition contains details for one aspect of the current
                     state of this API Resource. --- This struct is intended for direct
                     use as an array at the field path .status.conditions.  For example,
-                    type FooStatus struct{ // Represents the observations of a foo's
-                    current state. // Known .status.conditions.type are: \"Available\",
+                    \n type FooStatus struct{ // Represents the observations of a
+                    foo's current state. // Known .status.conditions.type are: \"Available\",
```
msherif1234, line 100: don't know why this is generated now with make manifest maybe something to do with k8s 1.25 update don't know

controllers/ingressnodefirewall_controller.go
```diff
@@ -271,6 +276,39 @@ func (r *IngressNodeFirewallReconciler) buildNodeStates(
 			return nil, err
 		}
 
+		// examine DeletionTimestamp to determine if object is under deletion
+		if firewallObj.ObjectMeta.DeletionTimestamp.IsZero() {
+			// The object is not being deleted, so if it does not have our finalizer,
+			// then lets add the finalizer and update the object. This is equivalent
+			// registering our finalizer.
+			if !stringInArray(ingressNodeFirewallFinalizerName, firewallObj.ObjectMeta.Finalizers) {
+				firewallObj.ObjectMeta.Finalizers = append(firewallObj.ObjectMeta.Finalizers, ingressNodeFirewallFinalizerName)
```
aojea, line 285: I think that you are mutating the object, you have to create a deepcopy with `firewallObjClone := firewallObj.DeepCopy` 

that may explain the races if you are mutating the cache objects

aojea: oh, this is kubebuilder, I don't know how it implements the Lists

controllers/ingressnodefirewall_controller.go
```diff
@@ -420,3 +458,25 @@ func mergeFirewallProtocolRules(a, b []infv1alpha1.IngressNodeFirewallProtocolRu
...
+		for _, item := range slice {
+			if item == s {
+				found = true
+				continue
+			}
+			result = append(result, item)
+		}
+	}
+	return
+}
```
aojea, line 482: this is more effective, just FYI
https://github.com/golang/go/wiki/SliceTricks#filter-in-place

controllers/ingressnodefirewall_controller.go
aojea: but it seems the `infList` is passed as parameter, you should DeepCopy to avoid mutating the list if someon else is using it

Makefile
```diff
@@ -125,7 +125,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(ENVTEST_ASSETS_DIR)/bin" go test ./... -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(ENVTEST_ASSETS_DIR)/bin" go test -race ./... -coverprofile cover.out
```
aojea, line 128: IIRC race and coverage have to be independent

msherif1234: like I have to run twice one for coverage and another to race ?

controllers/ingressnodefirewall_controller.go
```diff
@@ -272,7 +275,7 @@ func (r *IngressNodeFirewallReconciler) buildNodeStates(
 		}
 
 	withNextNode:
-		for _, node := range nodeList.Items {
+		for _, node := range nodeList.DeepCopy().Items {
```
aojea, line 278: you initialize nodeList in Line256 and populate it in Line272, I can't see another place where this can be mutated, do you need to deepcopy it?

msherif1234: you are right not needed here

aojea:
the changes LGTM to me, but I don't have the understanding of the codebase to understand if there are more things to check

martinkennelly (commented):
Small change otherwise LGTM

Makefile
```diff
@@ -127,6 +127,10 @@ vet: ## Run go vet against code.
 test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(ENVTEST_ASSETS_DIR)/bin" go test ./... -coverprofile cover.out
 
+.PHONY: test-race
+test-race: manifests generate fmt vet envtest ## Run tests and check for rance conditions.
```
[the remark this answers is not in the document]
msherif1234: dang it :(

PR #153, opened 2022-08-31
WIP: use RHEL9 images instead of RHEL8

msherif1234:


msherif1234:
RHCOS RHEL9 build isn't available yet and probably won't make it for 4.12 release

PR #158, opened 2022-09-06
Avoid mutating nodeState object directly to avoid race conditions

msherif1234:

### DIFF
--- a/controllers/ingressnodefirewall_controller.go
+++ b/controllers/ingressnodefirewall_controller.go
@@ -103,54 +103,55 @@ func (r *IngressNodeFirewallReconciler) Reconcile(ctx context.Context, req ctrl.
 			continue
 		}
 
+		nodeState := ingressNodeFirewallCurrentNodeState.DeepCopy()
 		// If the node name was found ...
 		// a) compare the specs or owner reference. If the specs or owner reference are different, update the current spec.
-		if !equality.Semantic.DeepEqual(ingressNodeFirewallCurrentNodeState.Spec, desiredNodeState.Spec) ||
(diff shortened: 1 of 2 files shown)


controllers/ingressnodefirewall_controller.go
```diff
@@ -103,6 +103,7 @@ func (r *IngressNodeFirewallReconciler) Reconcile(ctx context.Context, req ctrl.
 			continue
 		}
 
+		nodeState := ingressNodeFirewallCurrentNodeState.DeepCopy()
 		// If the node name was found ...
 		// a) compare the specs or owner reference. If the specs or owner reference are different, update the current spec.
 		if !equality.Semantic.DeepEqual(ingressNodeFirewallCurrentNodeState.Spec, desiredNodeState.Spec) ||
```
andreaskaris, line 109: I know it doesn't change anything to the correctness of the code, but let's make it "nodeState" everywhere then for consistency and readability even in the `if` clauses?
In line 109 right now we use `ingressNodeFirewallCurrentNodeState.Spec` for the comparison but then in line 135 `!equality.Semantic.DeepEqual(nodeState.Status, desiredNodeState.Status) `

controllers/ingressnodefirewall_controller.go
```diff
@@ -114,43 +115,43 @@ func (r *IngressNodeFirewallReconciler) Reconcile(ctx context.Context, req ctrl.
 				"req.Name", req.Name,
 				"ingressNodeFirewallCurrentNodeState.Name", ingressNodeFirewallCurrentNodeState.Name)
```
andreaskaris, line 116: `nodeState.Name`

andreaskaris (changes requested):
Tiny nit, see comments above.

PR #168, opened 2022-09-12
Fix image-references

locriandev:
Signed-off-by: D. Paolella <dpaolell@redhat.com>


msherif1234:
why merging to 4.12 instead of master ?

PR #172, opened 2022-09-15
Order overlap check should be done between different rules

### LINKED ISSUE #170: webhook overlapping order is broken
Change 3ea498b39cb489d91d0ad63f95636ecdad20aa8c seems to have broken overlapping ordering when there are two INFs with the exact same sourceCIDRs and they contain non-unique orders.

msherif1234:
within the same rule order is guranteed to be unique from API definition
fix #170 


martinkennelly:
/lgtm

Order for the existing INF is checked by func `orderIsUnique` and the change proprosed here just checks order for other INFs.

PR #174, opened 2022-09-15
label openshift-node-firewall namespace to avoid CSV failures during CI

msherif1234:


msherif1234:
doesn't seem to be needed when run the CI

PR #175, opened 2022-09-16
SDN-3488: Create golang package for libxdp

msherif1234:
info about libxdp https://man.archlinux.org/man/community/libxdp/libxdp.3.en
repo https://github.com/xdp-project/xdp-tools


dave-tucker (changes requested):
You're mixing 2 ebpf lifecycles still:

1. This load the program to the kernel and pins your maps - it's never attached
```
// Load pre-compiled programs into the kernel.
	objs := BpfObjects{}
	if err := LoadBpfObjects(&objs, &ebpf.CollectionOptions{Maps: ebpf.MapOptions{PinPath: pinDir}}); err != nil {
		return nil, fmt.Errorf("loading objects: pinDir:%s, err:%s", pinDir, err)
	}
```
2. This loads the program and creates the maps (again). This is the program that **does** get attached to the interface.
```
xdpProg := libxdploader.LoadXdpProgram(xdpProgName)
```

TL:DR the maps you are writing to reading/from are never used.

You'll need to use cilium/ebpfs query APIs to find your map FDs... or pass a pin location to libxdp somehow.

msherif1234:
> You're mixing 2 ebpf lifecycles still:
> 
> 1. This load the program to the kernel and pins your maps - it's never attached
> 
> ```
> // Load pre-compiled programs into the kernel.
> 	objs := BpfObjects{}
> 	if err := LoadBpfObjects(&objs, &ebpf.CollectionOptions{Maps: ebpf.MapOptions{PinPath: pinDir}}); err != nil {
> 		return nil, fmt.Errorf("loading objects: pinDir:%s, err:%s", pinDir, err)
> 	}
> ```
> 
> 2. This loads the program and creates the maps (again). This is the program that **does** get attached to the interface.
> 
> ```
> xdpProg := libxdploader.LoadXdpProgram(xdpProgName)
> ```
> 
> TL:DR the maps you are writing to reading/from are never used.
> 
> You'll need to use cilium/ebpfs query APIs to find your map FDs... or pass a pin location to libxdp somehow.

like `xdpProg := libxdploader.FindXdpProgramFromPin(infc.pinPath)` ?

dave-tucker:
> that was PoR is to let cilium handle the maps and just use libxdp as replacement of cilium's link but your observation is right and it explains why FW not taking effect after been attached, however for ur proposed options its not clear to me how to proceed further 

> (1) if we have the maps FDs from cilium how we can communicate that to libxdp which api ?

You can't as far as I can see. You'd instead use libxdp's `xdp_program__from_fd(int fd);` or `xdp_program__from_id(__u32 prog_id);`. Where either the program fd or id come from cilium/ebpf. See: https://pkg.go.dev/github.com/cilium/ebpf#Program.FD

>(2) I tried `xdpProg := libxdploader.FindXdpProgramFromPin(infc.pinPath)` where pinPath is where the maps are pinned to but that return ENOENT so probably this isn't right API

That's the wrong API - it's creating an `xdp_program` from a pinned file on bpffs so passing a pin that references a map won't work. If you find the pinned maps you use https://pkg.go.dev/github.com/cilium/ebpf#NewMapFromFD
Otherwise you'd have to find all loaded maps that are owned by the current process and use https://pkg.go.dev/github.com/cilium/ebpf#NewMapFromID - I'm not sure if cilium/ebpf has an API for the former though.

tohojo:
> You can't as far as I can see. You'd instead use libxdp's `xdp_program__from_fd(int fd);` or `xdp_program__from_id(__u32 prog_id);`.

Note that if you do this, you'll have to load the program as a `PROG_TYPE_EXT`, which means you'll also have to replicate all the feature probing that libxdp does to determine whether the dispatcher should be used or not. So I'm not sure if this is the best route.

You can go the other way, though:

1. Load program with `xdp_program__open_file()` followed by `xdp_program__attach()`.
2. Obtain fd for the loaded program from `xdp_program__fd()`
3. Feed fd to cilium-bpf's `NewProgramFromFD()`: https://pkg.go.dev/github.com/cilium/ebpf#NewProgramFromFD
4. Obtain the `ProgramInfo` object from `Program.Info`: https://pkg.go.dev/github.com/cilium/ebpf#Program.Info
5. Obtain the Map IDs from the `ProgramInfo` object: https://pkg.go.dev/github.com/cilium/ebpf#ProgramInfo.MapIDs
6. Obtain a map object with `NewMapFromID()`: https://pkg.go.dev/github.com/cilium/ebpf#NewMapFromID

And yeah, this is a bit hacky, but, well, hard to avoid that if you're mixing incompatible libraries :)

msherif1234:
> > You can't as far as I can see. You'd instead use libxdp's `xdp_program__from_fd(int fd);` or `xdp_program__from_id(__u32 prog_id);`.
> 
> Note that if you do this, you'll have to load the program as a `PROG_TYPE_EXT`, which means you'll also have to replicate all the feature probing that libxdp does to determine whether the dispatcher should be used or not. So I'm not sure if this is the best route.
This route led to a crash in libxdp with SIGSEGV
> 
> You can go the other way, though:
> 
> 1. Load program with `xdp_program__open_file()` followed by `xdp_program__attach()`.
> 2. Obtain fd for the loaded program from `xdp_program__fd()`
> 3. Feed fd to cilium-bpf's `NewProgramFromFD()`: https://pkg.go.dev/github.com/cilium/ebpf#NewProgramFromFD
> 4. Obtain the `ProgramInfo` object from `Program.Info`: https://pkg.go.dev/github.com/cilium/ebpf#Program.Info
> 5. Obtain the Map IDs from the `ProgramInfo` object: https://pkg.go.dev/github.com/cilium/ebpf#ProgramInfo.MapIDs
> 6. Obtain a map object with `NewMapFromID()`: https://pkg.go.dev/github.com/cilium/ebpf#NewMapFromID
Now what we will do with those newMaps
> 
> And yeah, this is a bit hacky, but, well, hard to avoid that if you're mixing incompatible libraries :)
yeah this horrible exercise for sure but lets see if we can get a working path and we still need to deal with packaging :)

tohojo:
> MM: now what we will do with those newMaps ?

Whatever it was you were doing with maps before - populating them, I suppose?

msherif1234:
@tohojo is there a way to consume cillium eBPF generated maps from pinPath ?

tohojo:
> @tohojo is there a way to consume cillium eBPF generated maps from pinPath ?

Hmm, you're already setting `LIBBPF_PIN_BY_NAME` on one of the maps, so you could set that on all of them and rely on libbpf auto-reuse of pinned maps when the program is loaded through libxdp. This would allow you to create and populate the maps using the cilium BPF library, and as long as they're pinned in the right directory, libxdp (through libbpf), *should* just pick this up and reuse the same maps.

To do this, you'll need to construct a `bpf_object_open_opts` struct (which comes from libbpf), set the `pin_root_path` member of this struct to the directory you're using (i.e., `/sys/fs/bpf/xdp_ingress_node_firewall_process`), and pass a pointer to this struct as the third argument to `xdp_program__open_file()`. I have absolutely no idea how to do this from the Go wrapper, though...

msherif1234:
> > @tohojo is there a way to consume cillium eBPF generated maps from pinPath ?
> 
> Hmm, you're already setting `LIBBPF_PIN_BY_NAME` on one of the maps, so you could set that on all of them and rely on libbpf auto-reuse of pinned maps when the program is loaded through libxdp. This would allow you to create and populate the maps using the cilium BPF library, and as long as they're pinned in the right directory, libxdp (through libbpf), _should_ just pick this up and reuse the same maps.
> 
> To do this, you'll need to construct a `bpf_object_open_opts` struct (which comes from libbpf), set the `pin_root_path` member of this struct to the directory you're using (i.e., `/sys/fs/bpf/xdp_ingress_node_firewall_process`), and pass a pointer to this struct as the third argument to `xdp_program__open_file()`. I have absolutely no idea how to do this from the Go wrapper, though...

ok I will explore this hopefully other map types allow me to pin them, I have seen example of C wrappers where it reference c structure fields so it seems possible I will play with option tomorrow that is probably the cleanest if we can make it to work Thanks @tohojo

tohojo:
> ok I will explore this hopefully other map types allow me to pin them, I have seen example of C wrappers where it reference c structure fields so it seems possible I will play with option tomorrow that is probably the cleanest if we can make it to work Thanks @tohojo

Awesome! You're welcome!

> while we are there if u can think of extensions to libxdp to smooth this intg with cilium you will get lots of future users just to leverage multi prog support leaving aside the RHEL support for now.

Not really? There's going to be an impedance match no matter what you do when trying to use two libraries at the same time. The only way to avoid this is to have native support for the libxdp protocol in the Cilium BPF library, I think. Did you have anything in particular in mind?

msherif1234:
@tohojo this last patch seems to work for me I just noticed I have to set the structure size otherwise I get too small opts error from libbpf
the other two maps one for events perf and others are status both are populated by the xdp prog and just read only from userspace I don't think we need to pin those two maps WDYT ?

tohojo:
> @tohojo this last patch seems to work for me I just noticed I have to set the structure size otherwise I get too small opts error from libbpf

Ah, very cool!

Did you look at the output of `xdp-loader status` after you've loaded the program? Just to check that everything looks right from the libxdp side...

> the other two maps one for events perf and others are status both are populated by the xdp prog and just read only from userspace I don't think we need to pin those two maps WDYT ?

It depends. What is "userspace" here? This is still the Go program, which is creating the maps using Cilium BPF, and then reading from them? And the XDP program is still writing (data or events) to them, right? Because in that case you still need to pin them for the XDP program to be able to pick them up and use the same maps as userspace. Otherwise, the XDP program will get separate copies of the maps, so you won't ever see any events or statistics when reading them from userspace.

pkg/libxdploader/README.md
```diff
@@ -0,0 +1,14 @@
+# Install libxdp headers
+```shell
+ git clone https://github.com/xdp-project/xdp-tools.git
+ cd xdp-tools
```
andreaskaris, line 4: We've got all of that stuff already packaged and ready to pull with yum (talking about something being supported, we support our own builds of libxdp and libbpf, and not builds from source):
~~~
yum install libxdp libxdp-devel xdp-tools libbpf libbpf-devel
~~~

I don't remember what rpm was required where, but we defo don't have to rebuild anything

pkg/libxdploader/libxdp_loader.go
```diff
@@ -0,0 +1,79 @@
+package libxdploader
```
andreaskaris, line 1: What about moving this into its own upstream repository and pulling it in, with the goal to make this a go goto library for libxdp Overkill?

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -199,22 +198,18 @@ func (infc *IngNodeFwController) IngressNodeFwAttach(ifacesName ...string) error
 		}
 
 		// Attach the program.
-		l, err := link.AttachXDP(link.XDPOptions{
-			Program:   objs.IngressNodeFirewallProcess,
-			Interface: int(ifID),
-		})
-		if err != nil {
-			errors = append(errors, fmt.Errorf("could not attach XDP program: %s", err))
+		xdpProName := "../ebpf/bpf_bpfel.o"
```
andreaskaris, line 201: nit for readability:
~~~
xdpProgName or xdpProgramName
xdpProg or xdpProgram
~~~

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -199,22 +198,18 @@ func (infc *IngNodeFwController) IngressNodeFwAttach(ifacesName ...string) error
...
+		xdpPro := libxdploader.LoadXdpProgram(xdpProName, infc.pinPath)
+		if xdpPro == nil {
+			errors = append(errors, fmt.Errorf("failed to load xdp prog %v", xdpPro))
```
andreaskaris, line 204: nit: `XDP program`

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -353,18 +353,22 @@ func (infc *IngNodeFwController) loadPinnedLinks() error {
...
-	if err := l.Unpin(); err != nil {
-		return fmt.Errorf("failed to unpin link for %s err: %q", ifName, err)
+	ifID, err := interfaces.GetInterfaceIndex(ifName)
+	if err != nil {
+		return err
 	}
-	if err := l.Close(); err != nil {
-		return fmt.Errorf("failed to close and detach link %s err: %q", ifName, err)
+
+	log.Printf("Running xdp program detach for  %v", p)
```
andreaskaris, line 365: nit: XDP

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -353,18 +353,22 @@ func (infc *IngNodeFwController) loadPinnedLinks() error {
 
 // cleanup will delete an interface's eBPF objects.
 func (infc *IngNodeFwController) cleanup(ifName string) error {
-	l, ok := infc.links[ifName]
+	p, ok := infc.progs[ifName]
 	if !ok {
-		return fmt.Errorf("failed to find Link object for interface %s", ifName)
+		return fmt.Errorf("failed to find xdp program object for interface %s", ifName)
```
andreaskaris, line 358: nit: XDP

Dockerfile.daemon
```diff
@@ -2,6 +2,27 @@ FROM golang:1.18 AS builder
 WORKDIR /go/src/github.com/openshift/ingress-node-firewall
 COPY . .
 
+#Install libxdp and libbpf headers
```
andreaskaris, line 5: Ah I see now why in the README earlier you mentioned how to build stuff.
The golang builder pod is debian based, so we have to compile  from source, then compile go and because it's all statically linked we can copy the binary over to a quay system. That creates a little bit of a dilemma, though, because we are not using a Red Hat supported version of libbpf or libxdp...?

Dockerfile
```diff
@@ -16,13 +16,36 @@ COPY controllers/ controllers/
...
+
+# Download libxdp source
+RUN wget -O xdp-tools.tar.gz "https://github.com/xdp-project/xdp-tools/releases/download/v${TOOLS_VERSION}/xdp-tools-${TOOLS_VERSION}.tar.gz" \
+    && tar xvfz xdp-tools.tar.gz \
+    && rm -f xdp-tools.tar.gz
+
+RUN cd $XDPTOOLS_SRC; ./configure; make install; cd $WORKDIR
+
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
```
andreaskaris, line 20: qq: only the daemon attached/detaches eBPF programs. Why does the operator need a libbpf / libxdp dependency?

Dockerfile.daemon
andreaskaris: With that said, we also have to update Dockerfile.daemon.openshift and there I suppose that we pull in the deps with yum?

Makefile
```diff
@@ -164,7 +164,7 @@ test-e2e: generate fmt vet manifests test-validation test-functional  ## Run e2e
 
 .PHONY: build
 build: prereqs generate fmt vet ## Build manager binary.
-	go build -o bin/manager main.go
+	CGO_ENABLED=1 go build -o bin/manager main.go
```
andreaskaris, line 167: Same question as before: why do the controllers need libxdp/libbpf when only the daemon actually manages attachments / detachments?

Dockerfile.daemon
msherif1234: correct or switch to use fedora builder so we can install those pkgs instead of the hack I have ?

pkg/ebpf/ingress_node_firewall_loader.go
msherif1234: my IDE suggested the stupid name I was using :(

Dockerfile.daemon
andreaskaris: Really weird: I can find the deps in brew but not in our rhel repos:
https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=2123985

~~~
yum install http://download.eng.bos.redhat.com/brewroot/vol/rhel-8/packages/xdp-tools/1.2.6/1.el8/x86_64/libxdp-1.2.6-1.el8.x86_64.rpm http://download.eng.bos.redhat.com/brewroot/vol/rhel-8/packages/xdp-tools/1.2.6/1.el8/x86_64/libxdp-devel-1.2.6-1.el8.x86_64.rpm http://download.eng.bos.redhat.com/brewroot/vol/rhel-8/packages/xdp-tools/1.2.6/1.el8/x86_64/xdp-tools-1.2.6-1.el8.x86_64.rpm
~~~

msherif1234: cc'd @tohojo

PR #191, opened 2022-09-22
OCPBUGS-1621: fix version to allow art build with the proper version with timestamp

msherif1234:
the logic in art.yaml wasn't able to find v0.0.1 in the daily build hence wasn't updating CSV correctly

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -3,5 +3,5 @@ package version
 var (
 	// Version contains the version number. This will be replaced by the linker at build
 	// time.
-	Version = "0.0.1"
+	Version = "4.12.0"
 )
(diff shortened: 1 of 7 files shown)


pkg/version/version.go
```diff
@@ -3,5 +3,5 @@ package version
 var (
 	// Version contains the version number. This will be replaced by the linker at build
```
andreaskaris, line 4: Sure we can do that ; ideally I'd still advocate for having it set through the linker at some point ... or we add a version.Commit field here that can be set to the commit's SHA through the linker (to make debugging a bit easier)

msherif1234: this will get us by https://issues.redhat.com/browse/OCPBUGS-1621
ART will replace the version with build TS so u will get something like 
```
oc get csv -n openshift-sriov-network-operator
sriov-network-operator.4.11.0-202209161807   SR-IOV Network Operator   4.11.0-202209161807
```

PR #195, opened 2022-09-30
User should be able to configure deny all rule for zero trust

### LINKED ISSUE #194: user should be able to configure 0.0.0.0/0 deny all rule
we should allow users to configure something like this
```yaml
apiVersion: ingressnodefirewall.openshift.io/v1alpha1
kind: IngressNodeFirewall
metadata:
  name: ingressnodefirewall-denyall
spec:
  interfaces:
  - eth0
  nodeSelector:
    matchLabels:
      do-node-ingress-firewall: 'true'
  ingress:
  - sourceCIDRs:
       - 0.0.0.0/0
    rules:
    - order: 20
      action: Deny
```
in case after configuring all the whitelist rules we need to drop everything else
user will be responsible to whitelist all ports their application will use in this case to ensure proper functionality

msherif1234:
Fix #194 while its risky but to ensure secure firewall users after defining all allow rules they wanted to drop everything else.
users need to be carefull to make sure all ports needed by their applications and protocols are added to the allow list to prevent any breakage.

changes passed e2e test
```sh

Ran 11 of 11 Specs in 173.373 seconds
SUCCESS! -- 11 Passed | 0 Failed | 0 Pending | 0 Skipped

```


msherif1234:
> Interesting change.. It would shorten the number of ingress rules required. I am trying to consider what at the risks here with introducing a protocol wildcard. Have you came up with any? I cannot think of any negatives right now. We should add e2e tests for this once my (#173 ) lands.

1- regarding ur comment "it would shorten # of rules" well depends if u are looking at from deny perspective then yes, however if u are using lots of openshift application "needs lots of pinholes" u will have to make sure to create allow rules for them before the deny all otherwise u will lose functionality.
2- the risk what I mentioned in 1 and in the issue for this PR if u aren't careful with using this u can either lock up ur cluster or lose functionality `validation webhook won't protect u in this case`
3- agree we need e2e coverage for that I didn't want to conflict with ur PR so defer this it can be done post FF.

now why adding this change? there is a debate going on about what is the `right` default action for infw so I wanted to be able to have both flavors 
- Users who want to block certain cidrs for known protocol and ports w/o need to do anything else those can configure things normally as we do.
- Users with zero trust then can terminate their rule with deny all and open all ports and protocol they need.

With that we can get users feedback which might help us in future release to decide on different default action ?

martinkennelly:
/lgtm
Sounds good. Thanks for your explanation.

PR #204, opened 2022-10-07
OCPBUGS-1825: fix daemonset crash when restart

msherif1234:
When DaemonSet restarts it was crashing when trying to attach and existing program with resource busy, we shouldn't treat this as an error just log a msg and continue

issue was reproduced on KinD and fix was verified there.


andreaskaris:
/lgtm

Cute, that check is back ;-)

PR #220, opened 2022-10-21
E2E test: only pay attention to internal IPs only and explicitly check for ipv6 address

martinkennelly:

Comment from @msherif1234 that wasn't addressed in #173


msherif1234:
yeah I called out that during E2E PR didn't double check the last diffs I assumed it was done :(
/lgtm
/approve

martinkennelly:
Tested it with KinD and `ip.To16()` acts as expected.

PR #221, opened 2022-10-24
E2E tests: add basic disruption tests

martinkennelly:

Expecting test named `IngressNodeFirewall policy is configurable after daemon deletion` to fail - seen with KinD and bare metal server.
INF policy configuration isn't working after daemon deletion.

Executed go mod tidy + go mod vendor
Executed make lint and its clean

### DIFF
--- a/controllers/ingressnodefirewallnodestate_controller.go
+++ b/controllers/ingressnodefirewallnodestate_controller.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 // IngressNodeFirewallNodeStateReconciler reconciles a IngressNodeFirewallNodeState object
@@ -37,6 +38,8 @@ type IngressNodeFirewallNodeStateReconciler struct {
 	Log       logr.Logger
 	Stats     *metrics.Statistics
 }
+
+var ingressNodeFirewallFinalizer = "ingressnodefirewall.openshift.io/finalizer"
 
 //+kubebuilder:rbac:groups=ingressnodefirewall.openshift.io,resources=ingressnodefirewallnodestates,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=ingressnodefirewall.openshift.io,namespace=ingress-node-firewall-system,resources=ingressnodefirewallnodestates,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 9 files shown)


martinkennelly:
P.S if you want me to add reboot test scenario LMK. I know you wanted to do it previously.

martinkennelly:
Also, I didnt hide these tests behind a flag because the tests arent adding that much latency to the overall test time. If they get too long later, we can do that.

martinkennelly:
/retest
Infra failure

test/e2e/functional/tests/e2e.go
```diff
@@ -837,8 +840,9 @@ var _ = Describe("Ingress Node Firewall", func() {
 			Expect(err).ShouldNot(HaveOccurred())
 			//wait for daemonset to be rolled out
 			infDaemonSet := &appsv1.DaemonSet{}
-			err = daemonset.WaitForDaemonSetReady(testclient.Client, infDaemonSet, OperatorNameSpace,
-				"ingress-node-firewall-daemon", retryInterval, timeout)
+			infDaemonSet.SetName("ingress-node-firewall-daemon")
```
msherif1234, line 843: Nit: pls use https://github.com/openshift/ingress-node-firewall/blob/master/test/consts/const.go#L15 for the ds name

controllers/ingressnodefirewallnodestate_controller.go
```diff
@@ -96,3 +121,7 @@ func (r *IngressNodeFirewallNodeStateReconciler) reconcileResource(
 	}
 	return ctrl.Result{}, nil
 }
+
+func isNodeStateDeletionInProgress(nodeState *infv1alpha1.IngressNodeFirewallNodeState) bool {
+	return !nodeState.ObjectMeta.DeletionTimestamp.IsZero()
+}
```
msherif1234, line 127: Hmm I did that in the past attempting to fix ut race but it really didn't help, I think this the right thing because fw and node state are different scope and there is an issue with garbage collection hence its recommended to use finalizers , the only thing here when I used the same there were issues I hit during testing which I don't remember them atm but given we have much more advanced e2e now and if this works I am fine with adding it

pkg/ebpfsyncer/ebpfsyncer.go
```diff
@@ -87,6 +87,12 @@ func (e *ebpfSingleton) SyncInterfaceIngressRules(
 		return err
 	}
 
+	// Ensure IngNodeFwController's pinned links and our managed interfaces align
+	// TODO: refactor to not have managed interfaces names from two sources
+	for _, linkName := range e.c.GetPinnedLinkNames() {
```
msherif1234, line 92: why we need this `createNewManager` will load pinned map ?

msherif1234: pls make sure with any changes to this package u run this manual ut from ebpfsyncer folder
`ebpfsyncer]$ go test -exec sudo ./...`

msherif1234 (commented):
I will take closer look to the new e2e test cases

controllers/ingressnodefirewallnodestate_controller.go
martinkennelly: If daemon is not listening for events when node state is deleted, it will miss cleaning up the node. Finalizer ensures the daemon consumes the delete event for node state, performs the cleanup ops, and then removes the finalizer.

msherif1234:
PR #354 will continue this work

PR #230, opened 2022-10-28
OCPBUGS-2934: unlink events unix socket during shutdown

msherif1234:
to reuse unix socket we have to unlink before attempting to reuse
when the events container restarts

To repro on kind we can restart the node container for example
```sh
 docker restart kind-worker
```

from the event container
```
 k logs -n ingress-node-firewall-system   ingress-node-firewall-daemon-pkfrd  -c events --previous
2022/10/28 21:19:59 Caught signal terminated: shutting down.
```

### DIFF
--- a/cmd/syslog/syslog.go
+++ b/cmd/syslog/syslog.go
@@ -3,6 +3,9 @@ package main
 import (
 	"fmt"
 	"log"
+	"os"
+	"os/signal"
+	"syscall"
 
 	"gopkg.in/mcuadros/go-syslog.v2"
 )
@@ -20,13 +23,40 @@ func main() {
 	server.SetFormat(syslog.RFC3164)
 	server.SetHandler(handler)
 
-	if err := server.ListenUnixgram(listenAddress); err != nil {
-		log.Fatal(err)
+	// check if syslog unix socket is already there then delete it.
+	if _, err := os.Stat(listenAddress); err == nil {
+		if err := os.Remove(listenAddress); err != nil {
+			log.Fatalf("error removing syslog socket %s: %s", listenAddress, err)
+		}
 	}
 
-	if err := server.Boot(); err != nil {
-		log.Fatal(err)
+	if err := server.ListenUnixgram(listenAddress); err != nil {
+		log.Fatalf("failed to listen to syslog unix socket %s: %s", listenAddress, err)
 	}
+
+	defer func() {
+		if err := os.Remove(listenAddress); err != nil {
+			log.Fatalf("error removing syslog socket %s: %s", listenAddress, err)
+		}
+	}()
+
+	if err := server.Boot(); err != nil {
+		log.Fatalf("failed to boot syslog server: %s", err)
+	}
+
+	// Unix sockets must be unlink()ed before being reused again.
+	// Handle common process-killing signals so we can gracefully shut down:
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, os.Interrupt, syscall.SIGTERM)
+	go func(c chan os.Signal) {
+		// Wait for a SIGINT or SIGKILL:
+		sig := <-c
+		log.Printf("Caught signal %s: shutting down.\n", sig)
+		// Stop listening (and unlink the socket if unix type):
+		if err := server.Kill(); err != nil {
+			log.Fatalf("failed to close server connections: %s", err)
+		}
+	}(sigc)
 
 	go func(channel syslog.LogPartsChannel) {
 		for logParts := range channel {


cmd/syslog/syslog.go
```diff
@@ -21,13 +24,27 @@ func main() {
...
+	// Handle common process-killing signals so we can gracefully shut down:
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, os.Interrupt, os.Kill, syscall.SIGTERM)
+	go func(c chan os.Signal) {
+		// Wait for a SIGINT or SIGKILL:
+		sig := <-c
+		log.Printf("Caught signal %s: shutting down.", sig)
+		// Stop listening (and unlink the socket if unix type):
+		_ = server.Kill()
+		_ = os.Remove(listenAddress)
```
frobware, line 44: I'd record the error if remove fails. Otherwise you'll be back where you started.

msherif1234: I was trying to make the signal handler very light weight

cmd/syslog/syslog.go
```diff
@@ -20,14 +23,37 @@ func main() {
...
+			log.Fatalf("error removing syslog socket %s: %s", listenAddress, err)
+		}
+	}
+
 	if err := server.ListenUnixgram(listenAddress); err != nil {
-		log.Fatal(err)
+		log.Fatalf("failed to listen to syslog unix socket %s, err %s", listenAddress, err)
 	}
 
+	defer os.Remove(listenAddress)
```
frobware, line 37: Also log any failure from the call to remove -- might help our future selves to understand any pattern where the file does not get unlinked cleanly.

cmd/syslog/syslog.go
```diff
@@ -20,14 +23,37 @@ func main() {
...
 	}
 
+	// Unix sockets must be unlink()ed before being reused again.
+	// Handle common process-killing signals so we can gracefully shut down:
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, os.Interrupt, syscall.SIGTERM)
+	go func(c chan os.Signal) {
+		// Wait for a SIGINT or SIGKILL:
+		sig := <-c
+		log.Printf("Caught signal %s: shutting down.", sig)
```
frobware, line 50: Add trailing `\n`

cmd/syslog/syslog.go
```diff
@@ -20,14 +23,41 @@ func main() {
...
+	// check if syslog unix socket is already there then delete it.
+	if _, err := os.Stat(listenAddress); err == nil {
+		if err := os.Remove(listenAddress); err != nil {
+			log.Fatalf("error removing syslog socket %s: %s", listenAddress, err)
+		}
+	}
+
 	if err := server.ListenUnixgram(listenAddress); err != nil {
-		log.Fatal(err)
+		log.Fatalf("failed to listen to syslog unix socket %s, err %s", listenAddress, err)
```
frobware, line 34: For consistency, format as `<msg>: <err>`

```go
log.Fatalf("failed to listen to syslog unix socket %s: %s", listenAddress, err)
```

cmd/syslog/syslog.go
```diff
@@ -20,14 +23,41 @@ func main() {
...
+	// Handle common process-killing signals so we can gracefully shut down:
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, os.Interrupt, syscall.SIGTERM)
+	go func(c chan os.Signal) {
+		// Wait for a SIGINT or SIGKILL:
+		sig := <-c
+		log.Printf("Caught signal %s: shutting down.\n", sig)
+		// Stop listening (and unlink the socket if unix type):
+		if err := server.Kill(); err != nil {
+			log.Printf("Failed to close server connections: %s.\n", err)
```
frobware, line 57: I think this should be fatal. You really want to exit. If the call to kill fails you'll still be stuck in `server.Wait()`.

cmd/syslog/syslog.go
```diff
@@ -20,14 +23,41 @@ func main() {
...
+
 	if err := server.Boot(); err != nil {
-		log.Fatal(err)
+		log.Fatalf("failed to boot syslog server: %s", err)
 	}
 
+	// Unix sockets must be unlink()ed before being reused again.
+	// Handle common process-killing signals so we can gracefully shut down:
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, os.Interrupt, syscall.SIGTERM)
```
martinkennelly, line 50: what about sigkill?

msherif1234: we were discussing tghat yesterday offline and it seems sigkil can't be caught or ignored hence I dropped it

PR #234, opened 2022-11-07
OCPBUGS-1435: added suggest ns to manifest

msherif1234:
Add suggested name space to manifest just to give users different options

### DIFF
--- a/bundle/manifests/ingress-node-firewall.clusterserviceversion.yaml
+++ b/bundle/manifests/ingress-node-firewall.clusterserviceversion.yaml
@@ -82,9 +82,13 @@ metadata:
         }
       ]
     capabilities: Basic Install
+    categories: Networking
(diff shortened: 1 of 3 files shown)


martinkennelly:
@msherif1234 Why isnt the namespace openshift-ingress-node-firewall?

msherif1234:
> @msherif1234 Why isnt the namespace openshift-ingress-node-firewall?
> 
> > Add suggested name space to manifest just to give users different options
> 
> Is this implying that ` operatorframework.io/suggested-namespace` is an alternative namespace suggestion? I tried to find docs for this and failed. Would appreciate if you have a link.

If you check the image in bug it doesn’t give you option to change ns this keyword apparently enable this fou f that out for similar pr got metallb operator

martinkennelly:
/lgtm
Thanks for the explanation.

PR #246, opened 2022-11-11
Revert "OCPBUGS-1435: added suggest ns to manifest"

msherif1234:
Reverts openshift/ingress-node-firewall#234


msherif1234:
no need to revert we can just rename ns in different PR

PR #262, opened 2022-11-29
SDN-3550: Add INFW XDP multi buffer support

msherif1234:
Multi buffer support is required to support
large MTU packets.

for reference on Multi buffer support pls check 
https://lore.kernel.org/bpf/cover.1634301224.git.lorenzo@kernel.org/
- Manual testing was done on KinD and it working fine.
- Manual CI with KinD passed for IPv4 and IPv6 test cases.
```bash

Ran 19 of 19 Specs in 253.835 seconds
SUCCESS! -- 19 Passed | 0 Failed | 0 Pending | 0 Skipped
```
- unit-test pass


msherif1234:
ci incorrectly build because I didn't commit the changes to ebpf auto generated pkg till I am able to build with the new changes
```bash
 deleted:    pkg/ebpf/bpf_bpfeb.o
 deleted:    pkg/ebpf/bpf_bpfel.o
```

msherif1234:
/hold
OCP CI failed JIT verification
```bash
17: (07) r3 += -72
	; if (unlikely(bpf_xdp_load_bytes(ctx, sizeof(struct ethhdr), (void *)&iph, sizeof(struct iphdr)) < 0)) {
	18: (bf) r1 = r6
	19: (b7) r2 = 14
	20: (b7) r4 = 20
	21: (85) call unknown#189
	invalid func unknown#189
	processed 22 insns (limit 1000000) max_states_per_insn 0 total_states 1 peak_states 1 mark_read 1
```
waiting on OCP build with RHEL9

msherif1234:
seeing the same ci issue reported in #296 not related to the new changes

dave-tucker:
@msherif1234 A couple of comments:

1. You need to change the section name to `SEC("xdp.frags")` since you've also pulled in libbpf > 1.0 and your section name is no longer compliant with their rules. That should be the only change required for multi-buf support.
2. You don't need to use the `bpf_xdp_load_bytes` helper unless you're accessing data after the first fragment which shouldn't be the case as you're only looking at L4 information.

Hopefully @tohojo can advise specifically on point 2.

msherif1234:
if I understand the `bpf_xdp_load_bytes` correctly if the offset with the same segment it will skip all the heavy lifting ? 
https://elixir.bootlin.com/linux/v5.19.17/source/net/core/filter.c#L3907

and yes @tohojo its expected to have jumpo frame for most deployments using this feature

tohojo:
On 13 March 2023 15:04:58 CET, "Mohamed S. Mahmoud" ***@***.***> wrote:
>if I understand the `bpf_xdp_load_bytes` correctly if the offset with the same segment it will skip all the heavy lifting ? 
>https://elixir.bootlin.com/linux/v5.19.17/source/net/core/filter.c#L3907

You'd think, but there's quite a bit of overhead anyway. Not sure why. See: ***@***.***/

So a simple before/after test would be good. Or, as Dave says, you could just keep using DPA...

msherif1234:
> On 13 March 2023 15:04:58 CET, "Mohamed S. Mahmoud" ***@***.***> wrote: if I understand the `bpf_xdp_load_bytes` correctly if the offset with the same segment it will skip all the heavy lifting ? https://elixir.bootlin.com/linux/v5.19.17/source/net/core/filter.c#L3907
> You'd think, but there's quite a bit of overhead anyway. Not sure why. See: ***@***.***/ So a simple before/after test would be good. Or, as Dave says, you could just keep using DPA...

at min I need `bpf_xdp_get_buff_len` for statistics update I will do before/after test to confirm Thank you all!!

msherif1234:
flaky ci
/retest

msherif1234:
using `bpf_ktime_get_ns` with kernel traces and bpftool to measure time to process ipv4 packets before and after:-

- before
```shell
 ping-61276   [010] d.s3.  3052.546589: bpf_trace_printk: ICMP packet rule(type:8, code:0) pkt(type:8, code 0)
            ping-61276   [010] d.s3.  3052.546590: bpf_trace_printk: Ingress node firewall process IPv4 packet time in ns 2278
            ping-61276   [010] d.s3.  3052.546610: bpf_trace_printk: Ingress node firewall action DENY -> XDP_DROP

```

- After
```shell
 <...>-41297   [011] d.s3.  1802.582266: bpf_trace_printk: ICMP packet rule(type:8, code:0) pkt(type:8, code 0)
 <...>-41297   [011] d.s3.  1802.582268: bpf_trace_printk: Ingress node firewall process IPv4 packet time in ns 6267
 <...>-41297   [011] d.s3.  1802.582306: bpf_trace_printk: Ingress node firewall action DENY -> XDP_DROP
```

tohojo:
> so this more than 30% performance hit that is very ugly is that because of the memcpy() ?

30%? I think you're off by an order of magnitude there - 2k->6k is more like 300% :open_mouth: 

As I mentioned above, I'm not actually sure where that performance hit is coming from, but cf the message I linked to above it's not entirely unexpected.

On all current drivers, you'll be fine to just keep using data and data_end for parsing the packet headers, even with a multibuf packet. It may be that some time in the future there will appear drivers that don't put the packet headers in the first fragment, but hopefully your tests will catch such a case before being deployed on such a driver. By that time we'd hopefully either have the performance regression solved upstream, or you can just move to dynptrs. So the easiest path forward here is probably just revert all the changes you've made to switch to `bpf_xdp_load_bytes()` :)

msherif1234:
> > so this more than 30% performance hit that is very ugly is that because of the memcpy() ?
> 
> 30%? I think you're off by an order of magnitude there - 2k->6k is more like 300% open_mouth
> 
> As I mentioned above, I'm not actually sure where that performance hit is coming from, but cf the message I linked to above it's not entirely unexpected.
> 
> On all current drivers, you'll be fine to just keep using data and data_end for parsing the packet headers, even with a multibuf packet. It may be that some time in the future there will appear drivers that don't put the packet headers in the first fragment, but hopefully your tests will catch such a case before being deployed on such a driver. By that time we'd hopefully either have the performance regression solved upstream, or you can just move to dynptrs. So the easiest path forward here is probably just revert all the changes you've made to switch to `bpf_xdp_load_bytes()` :)

so I implemented my own version of load bytes and used it everwhere, I wanted to know where the bottleneck is 
```C
__attribute__((__always_inline__)) static inline int
ingress_node_firewall_load_bytes(struct xdp_md *ctx, u32 offset, void *buf, u32 len) {
    void *data = (void *)(long)ctx->data;
    void *dataEnd = (void *)(long)ctx->data_end;
    void *dataStart = data + offset;

    if (unlikely(dataStart + len > dataEnd)) {
        return -1;
    }

    if (unlikely(buf == NULL || len == 0)) {
        return -1;
    }
    memcpy(buf, dataStart, len);
    return 0;
}
```

with that I am now seeing

```shell
           <...>-140577  [002] d.s3.  5386.740861: bpf_trace_printk: ICMP packet rule(type:8, code:0) pkt(type:8, code 0)
           <...>-140577  [002] d.s3.  5386.740863: bpf_trace_printk: Ingress node firewall process IPv4 packet took 4352
           <...>-140577  [002] d.s3.  5386.740873: bpf_trace_printk: Ingress node firewall action DENY -> XDP_DROP

```

so its not as bad as when used load bytes but still slow so `memcpy` for sure is not efficient

tohojo:
> so its not as bad as when used load bytes but still slow so `memcpy` for sure is not efficient.

Ah, cool, thank you for doing the test! I'm surprised it's that much just for `memcpy()`, but ah well...

> Do u think this is a good compromise till load bytes helper performance is fixed then I can just swap my local version with it w/o need to churn ? since we need to eventually use the helper function ?

Well, you don't need to move to it eventually, necessarily. You can use "dynptrs" instead, this was just merged for 6.4: https://lore.kernel.org/all/20230301154953.641654-1-joannelkoong@gmail.com/ - so I think it's probably better to just keep what you have and then move directly to dynptrs when that lands in a release? :)

msherif1234:
reverted back to DPA and make sure ci is passing on KinD, @tohojo @dave-tucker if no more comments pls LGTM
```shell

Ran 19 of 19 Specs in 606.767 seconds
SUCCESS! -- 19 Passed | 0 Failed | 0 Pending | 0 Skipped
```

bpf/ingress_node_firewall_kernel.c
```diff
@@ -417,20 +436,20 @@ ingress_node_firewall_main(struct xdp_md *ctx) {
...
     case DENY:
-        generate_event_and_update_statistics(ctx, (__u16)(dataEnd - data), DENY, ruleId, 1, ifId);
-        bpf_printk("Ingress node firewall action DENY -> XDP_DROP");
+        generate_event_and_update_statistics(ctx, bpf_xdp_get_buff_len(ctx), DENY, ruleId, 1, ifId);
+        ingress_node_firewall_printk("Ingress node firewall action DENY -> XDP_DROP");
         return XDP_DROP;
     case ALLOW:
-        generate_event_and_update_statistics(ctx, (__u16)(dataEnd - data), ALLOW, ruleId, 0, ifId);
-        bpf_printk("Ingress node firewall action ALLOW -> XDP_PASS");
+        generate_event_and_update_statistics(ctx, bpf_xdp_get_buff_len(ctx), ALLOW, ruleId, 0, ifId);
```
tohojo, line 443: The use of this helper will make the program fail to load on older kernels that don't have the helper available. Just want to double-check that this is OK?

msherif1234: Yes well aware of this

tohojo: Alright, cool. In that case, LGTM :)

tohojo (approved):

PR #325, opened 2023-03-31
fix typo in openshift config

msherif1234:
duplicate CRD entry for openshift config


msherif1234:
close it #326 will fix the typo as well as update CSV

PR #341, opened 2023-05-04
OCPBUGS-11885: Add ability to debug LPM lookups

msherif1234:

**- How to verify it**


martinkennelly (changes requested):
so previous to this change we updated the debug map always but following this change, its configurable right?

msherif1234:
> so previous to this change we updated the debug map always but following this change, its configurable right?

Correct to avoid performance impact

PR #393, opened 2023-09-02
Fix controller runtime package update from v0.15.1 to v0.16.1

msherif1234:

**- How to verify it**


cmd/daemon/daemon.go
```diff
@@ -15,24 +15,27 @@ package main
 
 import (
 	"flag"
+	"k8s.io/client-go/rest"
```
martinkennelly, line 18: format

cmd/daemon/daemon.go
```diff
@@ -15,24 +15,27 @@ package main
...
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	ingressnodefwiov1alpha1 "github.com/openshift/ingress-node-firewall/api/v1alpha1"
+	ingressnodefwv1alpha1 "github.com/openshift/ingress-node-firewall/api/v1alpha1"
+	"github.com/openshift/ingress-node-firewall/controllers"
+	"github.com/openshift/ingress-node-firewall/pkg/metrics"
+	"github.com/openshift/ingress-node-firewall/pkg/version"
 	"k8s.io/apimachinery/pkg/runtime"
```
martinkennelly, line 31: [stdlib]

[local]

[remote/k8s]

I think thats how we usually structure it.

main.go
```diff
@@ -86,8 +87,7 @@ func main() {
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                 scheme,
-		MetricsBindAddress:     metricsAddr,
-		Port:                   9443,
```
martinkennelly, line 90: where do we define the port now?

pkg/webhook/webhook_suite_test.go
```diff
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net"
 	"path/filepath"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
```
martinkennelly, line 25: format

main.go
msherif1234: from what I have seen its seems to be bundled with metricAddr I wish if they called `BindSocket` instead

msherif1234:
/retest
cluster didn't come up

main.go
msherif1234: I need to investigate further this port is used by webhook server 
```
	// Port is the port that the webhook server serves at.
	// It is used to set webhook.Server.Port if WebhookServer is not set.
	//
	// Deprecated: Use WebhookServer instead. A WebhookServer can be created via webhook.NewServer.
	Port int
```

msherif1234: fixed

PR #431, opened 2023-10-31
OCPBUGS-22756: update manifset operator image

msherif1234:
update operator image in the csv

### DIFF
--- a/bundle/manifests/ingress-node-firewall.clusterserviceversion.yaml
+++ b/bundle/manifests/ingress-node-firewall.clusterserviceversion.yaml
@@ -84,7 +84,7 @@ metadata:
     capabilities: Basic Install
     categories: Networking
     certified: "false"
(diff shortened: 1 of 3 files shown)


sosiouxme (commented):
ART is only looking at `/manifests/stable/ingress-node-firewall.clusterserviceversion.yaml`

and the important bit is that only references in `/manifests/stable/image-references` will get replaced (the replacement is by string, not semantic). so currently ART will replace `quay.io/openshift/origin-ingress-node-firewall:latest`; it won't replace `quay.io/openshift/ingress-node-firewall-operator:latest`

sosiouxme (approved):
/lgtm

ashwindasr:
I see that the test `ci/prow/ingress-node-firewall-e2e-metal-ipi ` succeeded but the PR isn't being updated. Issue with with github bot? :thinking:

PR #476, opened 2024-02-10
Use TCX hook as fall back if XDP cannot be attached successfully

msherif1234:
Today not all drivers can support large MTU for example AWS ENA driver can't handle MTU > 4000 and XDP will fail to attach with the following error
```sh
ens5: Failed to set xdp program, the current MTU (9001) is larger than the maximum allowed MTU (3498) while xdp is on
```

in recent kernel "6.6.0 and above" there is a new efficient hook called TCx so this PR add fall back mechanism such that if attaching to XDP fail for any reason TCX will be used as a fall back with logging message indicating this behavior

This PR implements https://issues.redhat.com/browse/RFE-5120
**- Dependancies**
requires RHEL 9.4 for TCX to work

**- How to verify it**
- e2e on fedora with forcing XDP failure 
- once we have OCP 4.16 build with RHEL 9.4 we will try the same on OCP cluster

using TCX ebpf hook as a fallback to do ingress node firewall in case XDP attaching fails


bpf/ingress_node_firewall_kernel.c
```diff
@@ -444,21 +444,83 @@ ingress_node_firewall_main(struct xdp_md *ctx) {
...
+        ingress_node_firewall_printk("TCX: Ingress node firewall process IPv4 packet");
+        result = ipv4_firewall_lookup(data, dataEnd, ifId);
+        break;
+    case bpf_htons(ETH_P_IPV6):
+        ingress_node_firewall_printk("TCX: Ingress node firewall process IPv6 packet");
+        result = ipv6_firewall_lookup(data, dataEnd, ifId);
+        break;
+    default:
+        ingress_node_firewall_printk("TCX: Ingress node firewall unknown L3 protocol TCX_NEXT");
+        return TCX_NEXT;
```
martinkennelly, line 495: If we have an unknown proto, what does `TCX_NEXT` mean? 
I see:
```
+/* (Simplified) user return codes for tcx prog type.
+ * A valid tcx program must return one of these defined values. All other
+ * return codes are reserved for future use. Must remain compatible with
+ * their TC_ACT_* counter-parts. For compatibility in behavior, unknown
+ * return codes are mapped to TCX_NEXT.
+ */
+enum tcx_action_base {
+	TCX_NEXT	= -1,
+	TCX_PASS	= 0,
+	TCX_DROP	= 2,
+	TCX_REDIRECT	= 7,
+};
```

So it returns unknown? And what happens then?

pkg/ebpf/bpf_bpfeb.go
```diff
@@ -91,7 +90,8 @@ type BpfSpecs struct {
 //
 // It can be passed ebpf.CollectionSpec.Assign.
 type BpfProgramSpecs struct {
-	IngressNodeFirewallProcess *ebpf.ProgramSpec `ebpf:"ingress_node_firewall_process"`
+	TcxIngressNodeFirewall        *ebpf.ProgramSpec `ebpf:"tcx_ingress_node_firewall"`
+	XdpIngressNodeFirewallProcess *ebpf.ProgramSpec `ebpf:"xdp_ingress_node_firewall_process"`
```
martinkennelly, line 94: nit: why "Process" at the end of `XdpIng.." ? better to keep the naming between tcx and xdp the same?

bpf/ingress_node_firewall_kernel.c
msherif1234: in case of chained TCX hooks it will go and run the next one if none it will be the same effect as `TCX_PASS`

pkg/ebpf/bpf_bpfeb.go
msherif1234: will do and keep the same naming

bpf/ingress_node_firewall_kernel.c
```diff
@@ -444,21 +444,83 @@ ingress_node_firewall_main(struct xdp_md *ctx) {
...
+        generate_event_and_update_statistics(ctx, ctx->len, DENY, ruleId, 1, ifId);
+        ingress_node_firewall_printk("TCX: Ingress node firewall action DENY -> TCX_DROP");
+        return TCX_DROP;
+    case ALLOW:
+        generate_event_and_update_statistics(ctx, ctx->len, ALLOW, ruleId, 0, ifId);
+        ingress_node_firewall_printk("TCX: Ingress node firewall action ALLOW -> TCX_NEXT");
+        return TCX_NEXT;
+    default:
+        ingress_node_firewall_printk("TCX: Ingress node firewall action UNDEF");
+        return TCX_NEXT;
```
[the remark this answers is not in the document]
msherif1234: yes if no match the default action is allow

PR #487, opened 2024-03-07
OCPBUGS-30561: Set infra labels in ingress-node-firewall-operator

msherif1234:
set infra annotation labels for INFW manifest

### DIFF
--- a/bundle/manifests/ingress-node-firewall.clusterserviceversion.yaml
+++ b/bundle/manifests/ingress-node-firewall.clusterserviceversion.yaml
@@ -86,6 +86,13 @@ metadata:
     certified: "false"
     containerImage: quay.io/openshift/origin-ingress-node-firewall:latest
     createdAt: "2022-10-28 00:00:00"
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "true"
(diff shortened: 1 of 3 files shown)


bundle/manifests/ingress-node-firewall.clusterserviceversion.yaml
```diff
@@ -86,6 +86,13 @@ metadata:
     certified: "false"
     containerImage: quay.io/openshift/origin-ingress-node-firewall:latest
     createdAt: "2022-10-28 00:00:00"
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "false"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
```
martinkennelly, line 95: are there more platforms we have to think about? alibaba, etc? do youve a link to where these labels are defined?

martinkennelly: never mind, i see it linked in the bug.

PR #510, opened 2024-04-12
Add ebpf C linter

msherif1234:
remove deperceated pkg and update golanglint-ci
and ofcourse fixed new lint errors/warnings
recompiled ebpf files after new fmt changes


msherif1234:
needs https://github.com/openshift/release/pull/50950

controllers/ingressnodefirewallconfig_controller_test.go
```diff
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"k8s.io/utils/ptr"
```
frobware, line 6: Reorder to have standard library imports first, then k8s? It seems it was moved from a previous "k8s" grouping.

go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/openshift/ingress-node-firewall
 
-go 1.21
+go 1.21.0
```
frobware, line 3: I've seen this happen on a few go mod changes recently. If you have `toolchain` do you actually need to be so specific with `go 1.21.0` or would `go 1.21` suffice?

pkg/interfaces/interfaces_test.go
```diff
@@ -81,7 +81,9 @@ func TestGetInterfaceIndices(t *testing.T) {
 		t.Fatal(err)
 	}
 	t.Logf("bond interface index %d", bond.Attrs().Index)
-	defer netlink.LinkDel(bond)
+	defer func() {
+		_ = netlink.LinkDel(bond)
+	}()
```
frobware, line 86: If this is purely a stylistic change which continues to ignore the error from `LinkDel` is there actually any value?

test/e2e/namespaces/namespaces.go
```diff
@@ -63,7 +63,7 @@ func CleanPods(namespace string, cs *testclient.ClientSet) error {
 		return nil
 	}
 	err := cs.Pods(namespace).DeleteCollection(context.Background(), metav1.DeleteOptions{
-		GracePeriodSeconds: pointer.Int64Ptr(0),
+		GracePeriodSeconds: ptr.To[int64](0),
```
frobware, line 66: Ooh! Generics! :)

frobware (changes requested):
Generally LGTM. Why did the two eBPF object files change?

go.mod
msherif1234: `make vendor` is the one changing it for me

pkg/interfaces/interfaces_test.go
msherif1234: wanted to keep linter quiet but I will add check and err log here while I am here

msherif1234:
> Generally LGTM. Why did the two eBPF object files change?

I guess clang isn't smart enough to realize the diffs are all just white space

frobware:
> Generally LGTM. Why did the two eBPF object files change?

... via `make ebpf-generate`.

pkg/ebpfsyncer/ebpfsyncer.go
```diff
@@ -92,7 +92,7 @@ func (e *ebpfSingleton) SyncInterfaceIngressRules(
 		// Wait for a SIGTERM
 		<-c
 		if e.c != nil {
-			e.resetAll()
+			_ = e.resetAll()
```
frobware, line 95: Should any return value here be logged?

msherif1234: not really we are handling `SIGTERM` the container is going away at this point so its best effort cleanup `resetAll` will generate enough logs for error cases

PR #518, opened 2024-04-30
update cilium to version 0.15.0

msherif1234:
update cilium to v0.15.0

this PR update cilium pkg as dependabot atm is stuck because of the dependency on go1.22 once that dependency is addressed dependabot will be doing this update in the future

go1.22 is blocked by https://github.com/openshift-metal3/dev-scripts/pull/1645


hack/generators.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM fedora:37
+FROM fedora:40
```
frobware, line 1: Is the bump here required because of the Cilium bump?

msherif1234: no this is independent but while I was updating the cilium pkg I updated the os to pickup newer version of clang compiler

frobware: Does cilium 0.15 require a newer version of clang? I was just trying to understand whether the Fedora 37->40 bump was a necessity.

msherif1234: no it doesn't both updates are independent I just combined to avoid the another PR overhead

frobware: /lgtm

PR #519, opened 2024-05-03
Users prefer to have flexibility configuring SSH firewall rules

msherif1234:

Currently the SSH is protected by failsafe webhook preventing any attempts to configure any rules
for SSH this PR remove SSH from failsafe rules

**- How to verify it**
passed e2e and ut tests

remove SSH from failsafe rules to give user flexibility to configure SSH rules


msherif1234:
> /lgtm Makes sense. Is there documentation we need to update for OCP?

will check with doc team Thanks!!

PR #527, opened 2024-05-29
OCPBUGS-34380: fix ingress node firewall lookup on s390x "bigendian" arch

msherif1234:
fix INFW key format on bigendian arch

**- How to verify it**
- test on s390x arch
- pass ut and e2e testing

fix ingress node firewall ebpf lookup on s390x arch


djslavens (approved):
/lgtm

PR #530, opened 2024-06-06
fix go version to get successfull ART build

msherif1234:
Image build with the following error
`Required/recommended Go toolchain version \'1.22.2\' is not supported yet.`


msherif1234:
ci got broken with authentication error which is fixed in main but missing in 4.16

PR #572, opened 2024-10-24
SDN-5383: enable INFW tcx hook with bpfman

msherif1234:
Enable INFW TCX hook with the updated bpfman operator version


anfredette (approved):
/lgtm

PR #651, opened 2025-07-16
OCPBUGS-58577: bump github.com/golang/glog-1.0.0 to 1.2.4

MrSanketkumar:
backport : #650 

Bumping the glog to version 1.2.4 introduces a dependency on go-cmp, which has also been updated to v0.6.0. I’ve checked version 4.14 and earlier, and they all use the same version of go-cmp, which is included in glog-1.2.4.

**- How to verify it**


MrSanketkumar:
@sdodson The 4.14 tracker which is dependent on 4.13 is not yet verified due to build is not available, that's why we are getting Jira invalid. 

Right now we just want to override the failed prow job and once 4.14 build is available with the PR change we can verify and then it can be merged

PR #720, opened 2026-06-18
CORENET-7108: Implement TLS profile compliance

tpantelis:
See individual commits for details.

Overview:
- Use the centralized OCP TLS profile for the operator webhook and metrics servers
- Use the centralized OCP TLS profile for the daemonset `kube-rbac-proxy` via templating
- Removed the `kube-rbac-proxy` side car for the operator in lieu of using controller-runtime's native support for TLS and authorization. This eliminates having to deal with the hard-coded --tls args in the embedded bash script and restarting the container when the TLS profile is updated.

## Summary

* **New Features**
  * Enabled TLS profile support for node firewall traffic, including dynamic TLS min-version and cipher suite selection (with safe fallback when unset).

* **Security & Maintenance**
  * Metrics are now served over **HTTPS** on **:9300** using a required metrics certificate directory; removed the prior metrics proxy startup wrapper.
  * Updated RBAC for TLS profile/API server reads and auth review capabilities.
  * Migrated webhooks to typed handlers and expanded TLS-focused test coverage.
  * Upgraded dependencies and operator build tooling.


tpantelis:
@coderabbitai Why was your review skipped? You indicated "An error occurred during the review process. Please try again later." in comment https://github.com/openshift/ingress-node-firewall/pull/720#issuecomment-4743214043. What was the error?

tpantelis:
> I can re-run the full review now if you'd like.

@coderabbitai Please re-run the full review

controllers/ingressnodefirewallconfig_controller.go
```diff
@@ -160,6 +163,17 @@ func (r *IngressNodeFirewallConfigReconciler) syncIngressNodeFwConfigResources(c
 		}
 	}
 
+	// Add TLS configuration for kube-rbac-proxy
+	// Always set these keys (even if empty) so the template doesn't error on missing keys
+	tlsProfileSpec := r.GetTLSProfileSpec()
```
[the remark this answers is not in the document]
tpantelis: No need to check for nil. The one call site always sets it.

config/manager/manager.yaml
```diff
@@ -31,11 +31,16 @@ spec:
         - /manager
         args:
           - "--health-probe-bind-address=:8081"
-          - "--metrics-bind-address=127.0.0.1:39400"
+          - "--metrics-bind-address=:9300"
+          - "--metrics-cert-dir=/etc/pki/tls/metrics-certs"
```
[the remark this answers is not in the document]
tpantelis: The metrics endpoint is authenticated via controller-runtime's `filters.WithAuthenticationAndAuthorization` (main.go line 135). This performs TokenReview-based authentication and SubjectAccessReview-based authorization exactly like `kube-rbac-proxy`.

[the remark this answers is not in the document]
tpantelis: @coderabbitai you good?

controllers/suite_test.go
```diff
@@ -54,6 +56,7 @@ const (
 var cfg *rest.Config
 var k8sClient client.Client
 var testEnv *envtest.Environment
+var tlsProfileSpec atomic.Value // holds *configv1.TLSProfileSpec
```
jcaamano, line 60: Nit

I wonder about choice of an atomic value. Nothing wrong but there should not be any concurrency over it: either tests run serially or in their own process when paralell. Maybe I am missing something that would be good for me to know.

pkg/tls/profile.go
```diff
@@ -0,0 +1,132 @@
...
+		return fmt.Errorf("failed to setup TLS profile watcher: %w", err)
+	}
+
+	return nil
+}
+
+// ConvertCiphersToIANA converts a list of cipher suites from OpenSSL names to IANA names.
+// OpenShift TLS profiles use OpenSSL cipher names, but Go's crypto/tls and kube-rbac-proxy
+// expect IANA names.
+func ConvertCiphersToIANA(ciphers []string) []string {
```
jcaamano, line 123: Maybe something github.com/openshift/controller-runtime-common/pkg/tls should provide

main.go
```diff
@@ -160,6 +192,12 @@ func main() {
 		Log:          ctrl.Log.WithName("controllers").WithName("IngressNodeFirewallConfig"),
 		Namespace:    nameSpace,
 		PlatformInfo: platformInfo,
+		GetTLSProfileSpec: func() *configv1.TLSProfileSpec {
+			if !crypto.ShouldHonorClusterTLSProfile(tlsProfile.Adherence) {
+				return nil
+			}
```
jcaamano: One bit of confusion that I have here is that `ShouldHonorClusterTLSProfile` is also called within `GetProfileInfo` and even when `false`, `tlsProfile.ProfileSpec` is set. But here we seem to be overriding that returning `nil`.

Wouldn't it make more sense to have `GetProfileInfo` return `nil` and check for nilness here?

config/manager/manager.yaml
```diff
@@ -57,56 +62,15 @@ spec:
...
           - name: ingress-node-firewall-controller-manager-metrics-certs
             mountPath: /etc/pki/tls/metrics-certs
-            readOnly: True
+            readOnly: true
       volumes:
         - name: ingress-node-firewall-controller-manager-metrics-certs
           secret:
             secretName: ingress-node-firewall-controller-manager-metrics-certs
-            optional: true
+            optional: false
```
jcaamano, line 74: Should this have been non-optional all along, and not specific to this PR?

jcaamano (commented):
Minor questions, but does LGTM

controllers/suite_test.go
tpantelis: It's not really about concurrency but it's still accessed by multiple threads so we need memory barrier synchronization to ensure writes are visible across threads (and Go would flag a data race without synchronization).

pkg/tls/profile.go
tpantelis: yeah it has `OpenSSLToIANACipherSuites` but, unfortunately, that doesn't  check if a name is already IANA.

main.go
tpantelis: I had that originally but we also need the `tlsProfile.ProfileSpec` for the `SetupProfileWatch` in L216. I refactored it so `profileSpec` and `adherence` are private and `SetupProfileWatch` is a method which is cleaner.

config/manager/manager.yaml
tpantelis: yeah it could've been and the bash code to wait for certs mounted wouldn't have been necessary. I see inconsistency across the repos with this, ie some set `optional` to true and wait for certs and some set it to false and don't wait. But we've removed the bash script and controller runtime handles missing cert file on startup.

main.go
jcaamano: I don't understand or know how to interpret `TLSAdherencePolicy` and `ShouldHonorClusterTLSProfile`.

If we are ready to comply with the cluster wide profile, why would we ever fallback to something else?

jcaamano: Also on the fallback for the manager, wouldn't we be losing the configuration we had to limit the cipher suites? Should we hardcode a TLSOpt for it? Or maybe the defaults the controller runtime is using are aligned with what we had?

tpantelis: > I don't understand or know how to interpret TLSAdherencePolicy and ShouldHonorClusterTLSProfile

The `TLSAdherencePolicy` provides a mechanism for backwards compatibility. It prevents breaking changes during OCP upgrades while allowing cluster admins to opt into strict TLS compliance when they're ready. With `LegacyAdheringComponentsOnly` (the default now), components that do not already honor the centralized TLS profile should continue to use their individual TLS configurations. With `StrictAllComponents`, all components must honor the cluster-wide TLS profile.

> Also on the fallback for the manager, wouldn't we be losing the configuration we had to limit the cipher suites?

So those cipher suites were for the `kube-rbac-proxy` and the manager uses Go's crypto/tls with TLS 1.2 by default which includes those cipher suites. So Go's defaults are well-vetted and are at least equivalent to (or better than) what `kube-rbac-proxy` had configured.

jcaamano: I get now, thanks. It's confusing because `already` is a relative timeline. Relative to what?  I guess it means that if you are looking into `TLSAdherencePolicy` you are not a component that already honored the the centralized TLS profile in the first place.

tpantelis: yeah that part can get a little confusing since the TLS profile stuff is relatively new. Basically, the way I look at it, if you have a new component added in 5.0+ that has no prior legacy behavior then you can/should ignore the `TLSAdherencePolicy` and always apply the central TLS profile.

PR #730, opened 2026-07-09
OCPBUGS-67363: fixes ip traffic fragmentation verification

raphaelvrosa:
Impact: Attackers can access blocked ports by fragmenting IP packets in specific ways that exploit the firewall's failure to validate IP fragmentation headers before extracting Layer 4 port information.

Root Cause: The eBPF code assumes all packets have complete TCP/UDP/SCTP headers immediately following the IP header, without checking if the packet is a non-first fragment that contains only payload data.

Solution:
- Check iph->frag_off before extracting L4 headers
- Deny all non-first fragments (secure by default)
- First fragments (offset = 0) still processed normally with L4 inspection

The solution was verified and validate locally with kind and ovn-kubernetes as CNI. The ebpf program was loaded, and traffic was blocked.
End to end tests with hping3 are added to validate the solution.

**- How to verify it**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved firewall handling of fragmented IPv4 traffic by denying non-initial fragments that lack transport-layer information.
  * Added coverage to verify fragmented TCP traffic cannot bypass blocked-port rules across supported network protocols and IP versions.

* **Chores**
  * Updated the bundled cert-manager version for installation and local cluster cleanup workflows.
  * Enhanced network testing tools to support fragmented traffic validation.


bpf/ingress_node_firewall_kernel.c
```diff
@@ -291,8 +306,13 @@ ipv6_firewall_lookup(void *data, void *data_end, __u32 ifId) {
...
+  int l4_result = ip_extract_l4info(data, data_end, &proto, &dstPort, &icmpType,
+                                     &icmpCode, 0);
+  if (unlikely(l4_result < 0)) {
+    if (l4_result == -2) {
+      // Non-first IP fragment: deny by security policy
+      return SET_ACTIONRULE_RESPONSE(DENY, INVALID_RULE_ID);
+    }
     ingress_node_firewall_printk("failed to extract l4 info");
     return SET_ACTION(UNDEF);
   }
```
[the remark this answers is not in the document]
raphaelvrosa: @danwinship should this PR address this?

test/e2e/transport/transport.go
```diff
@@ -224,3 +229,62 @@ func ncClientTransport(client *testclient.ClientSet, sourcePod *corev1.Pod, sour
...
+	return hpingClient(client, sourcePod, destinationIP, dPort, "--udp", "-f", "-c", "1")
+}
+
+func hpingUDPFragV6(client *testclient.ClientSet, sourcePod *corev1.Pod, destinationIP, dPort string) (string, string, error) {
+	return hpingClient(client, sourcePod, destinationIP, dPort, "--udp", "-f", "-c", "1", "-6")
+}
+
+func hpingClient(client *testclient.ClientSet, sourcePod *corev1.Pod, destinationIP, dPort string, flags ...string) (string, string, error) {
+	command := []string{"sh", "-c", fmt.Sprintf("hping3 %s -p %s %s 2>&1 || true", strings.Join(flags, " "), dPort, destinationIP)}
+	return exec.RunExecCommand(client, sourcePod, command...)
```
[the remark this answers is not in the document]
raphaelvrosa: Done

test/e2e/functional/tests/e2e.go
```diff
@@ -1620,6 +1669,43 @@ func reachabilityCheck(reach reachable, podNameObj map[string]*corev1.Pod,
...
+func reachabilityCheckFragmentation(reach reachable, podNameObj map[string]*corev1.Pod,
+	protocols []ingressnodefwv1alpha1.IngressNodeFirewallRuleProtocolType) {
+	sourcePod := podNameObj[reach.source]
+	for _, protocol := range protocols {
+		if skipProtocol(protocol, !v6Enabled) {
+			continue
+		}
+		if !infwutils.IsTransportProtocol(protocol) {
+			continue
+		}
```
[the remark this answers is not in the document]
raphaelvrosa: Done.

test/e2e/transport/transport.go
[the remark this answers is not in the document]
raphaelvrosa: Changes done based on hping3 returned exist code:
Exit 0 → hping3 got a response (packet reached target) — no error, parsed by IsHpingResponseSeen
Exit 1 → no response (dropped/filtered) — expected "denied" path, allowed through
Any other error (127 missing binary, 2 bad args, permission denied, exec failure) → Fail() immediately

test/Dockerfile.netcat
```diff
@@ -1,4 +1,5 @@
 FROM alpine:latest
 
-RUN apk add --no-cache --update --verbose bash nmap-ncat && \
+RUN apk add --no-cache --update --verbose bash nmap-ncat hping3 \
+    --repository=https://dl-cdn.alpinelinux.org/alpine/edge/testing && \
```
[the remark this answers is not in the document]
raphaelvrosa: Changes done to have hping3 install in different repository.
This is an image only used for local tests, keeping it latest is fine.

config/manager/kustomization.yaml
```diff
@@ -9,7 +9,7 @@ patchesStrategicMerge:
 - manager_webhook_patch.yaml
 images:
 - name: controller
-  newName: quay.io/openshift/origin-ingress-node-firewall
-  newTag: latest
+  newName: localhost/ingress-node-firewall-controller
+  newTag: dev
```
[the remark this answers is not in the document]
raphaelvrosa: Fixed.

test/e2e/transport/transport.go
```diff
@@ -224,3 +229,62 @@ func ncClientTransport(client *testclient.ClientSet, sourcePod *corev1.Pod, sour
...
+	return hpingClient(client, sourcePod, destinationIP, dPort, "--udp", "-f", "-c", "1")
+}
+
+func hpingUDPFragV6(client *testclient.ClientSet, sourcePod *corev1.Pod, destinationIP, dPort string) (string, string, error) {
+	return hpingClient(client, sourcePod, destinationIP, dPort, "--udp", "-f", "-c", "1", "-6")
+}
+
+func hpingClient(client *testclient.ClientSet, sourcePod *corev1.Pod, destinationIP, dPort string, flags ...string) (string, string, error) {
+	command := []string{"sh", "-c", fmt.Sprintf("hping3 %s -p %s %s 2>&1", strings.Join(flags, " "), dPort, destinationIP)}
+	return exec.RunExecCommand(client, sourcePod, command...)
```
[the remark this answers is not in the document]
raphaelvrosa: Fixed.

test/e2e/transport/transport.go
```diff
@@ -224,3 +229,62 @@ func ncClientTransport(client *testclient.ClientSet, sourcePod *corev1.Pod, sour
...
+		}
+		return hpingTCPFragV4(client, sourcePod, destinationIP, destinationPort)
+	case ingressnodefwv1alpha1.ProtocolTypeUDP:
+		if v6 {
+			return hpingUDPFragV6(client, sourcePod, destinationIP, destinationPort)
+		}
+		return hpingUDPFragV4(client, sourcePod, destinationIP, destinationPort)
+	case ingressnodefwv1alpha1.ProtocolTypeSCTP:
+		// hping3 does not support SCTP
+		return "", "", nil
```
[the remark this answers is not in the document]
raphaelvrosa: Fixed.

bpf/ingress_node_firewall_kernel.c
```diff
@@ -110,6 +111,15 @@ ip_extract_l4info(void *data, void *dataEnd, __u8 *proto, __u16 *dstPort,
...
     *proto = iph->protocol;
+
+    // Security: Check for IP fragmentation (RFC 791)
+    // Non-first fragments don't have L4 headers, deny them to prevent bypass
+    __u16 frag_off = bpf_ntohs(iph->frag_off);
+    if (unlikely(frag_off & IP_OFFSET_MASK)) {
+      // Fragment offset > 0: this is a non-first fragment
+      // No L4 header present, deny by policy
+      return -2;
+    }
```
danwinship, line 118: If we're going to deny non-first fragments, it seems like we should deny first fragments too, since without the followup fragments the initial fragment will be useless.

So something like `unlikely((frag_off & IP_OFFSET_MASK) != 0 || (frag_off & IP_MF))`

bpf/ingress_node_firewall_kernel.c
```diff
@@ -97,6 +97,7 @@ volatile const __u32 debug_lookup = 0;
  * Return:
  * 0 for Success.
  * -1 for Failure.
+ * -2 for non-first IP fragment (security: deny fragments without L4 headers).
```
danwinship: Should probably have `#defined` return values.

And this should be clearer about what the different statuses imply:
- ` 0`: success (extracted L4 info)
- `-1`: packet is truncated; allow it to continue to the kernel, where it should be rejected
- `-2`: packet is fragmented; reject and do not pass to kernel (because INF does not support reassembling and checking fragmented packets, so we have to assume all fragmented packets are denied).

bpf/ingress_node_firewall_kernel.c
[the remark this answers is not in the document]
danwinship: no, we'll do a followup PR

test/e2e/functional/tests/e2e.go
```diff
@@ -828,6 +828,43 @@ var _ = Describe("Ingress Node Firewall", func() {
 					},
 				},
 			},
+			{
+				"should deny fragmented TCP packets to blocked port",
```
danwinship: This is just a dup of the `"block a port with a single rule defining the destinations port"` case isn't it?

test/e2e/functional/tests/e2e.go
```diff
@@ -973,6 +1010,11 @@ var _ = Describe("Ingress Node Firewall", func() {
 					reach := reach
 					reachabilityCheck(reach, podNameObj, entry.protocols, reach.connectivity)
 				}
+				// test fragmented packets are always denied
+				for _, reach := range entry.reachables {
+					reach := reach
```
danwinship: the `reach := reach` is not needed with go >= 1.22

test/e2e/functional/tests/e2e.go
```diff
@@ -1515,7 +1557,17 @@ func isConnectivitySeen(client *testclient.ClientSet, protocol ingressnodefwv1al
 		// Connectivity will be confirmed with server output later and expecting server output to contain clients IP.
 		_, _, _ = transport.ConnectToPortFromPod(client, protocol, v6, sourcePod, sourceIP, destinationIP, destinationPort)
 		serverResult := <-serverResultsCh
-		return strings.Contains(serverResult, sourceIP)
+		ncatResult := strings.Contains(serverResult, sourceIP)
+
+		// hping3 fragmented connectivity test (runs side by side with ncat)
```
danwinship: I would expect "side by side" to mean that they run in parallel, but that's not what's happening here. (Also, it's not clear what "ncat" refers to since nothing else in this file mentions "ncat"... if you mean the `transport.ConnectToPortFromPod` test, say that.)

test/e2e/functional/tests/e2e.go
```diff
@@ -1515,7 +1557,17 @@ func isConnectivitySeen(client *testclient.ClientSet, protocol ingressnodefwv1al
...
+		hpingStdout, hpingStderr, hpingErr := transport.HpingFragmentedConnect(client, protocol, v6, sourcePod, destinationIP, destinationPort)
+		if hpingErr != nil && !strings.Contains(hpingErr.Error(), "exit code 1") {
+			log.Printf("hping3 frag probe failed to execute: %v", hpingErr)
+		}
+		hpingResult := transport.IsHpingResponseSeen(hpingStdout, hpingStderr)
```
danwinship: I assume the `!strings.Contains(hpingErr.Error(), "exit code 1")` is because that error code means "failed to connect"? It's weird to have _some_ of the logic for parsing the return value here, and _some_ in `IsHpingResponseSeen`. It would be better to just have `transport.HpingFragmentedConnect` parse the return value itself and return `(connected bool, err error)`

test/e2e/functional/tests/e2e.go
```diff
@@ -973,6 +1010,11 @@ var _ = Describe("Ingress Node Firewall", func() {
 					reach := reach
 					reachabilityCheck(reach, podNameObj, entry.protocols, reach.connectivity)
 				}
+				// test fragmented packets are always denied
+				for _, reach := range entry.reachables {
+					reach := reach
+					reachabilityCheckFragmentation(reach, podNameObj, entry.protocols)
+				}
```
danwinship: why is this loop needed if you also check this from `isConnectivitySeen`?

test/e2e/transport/transport.go
```diff
@@ -224,3 +229,62 @@ func ncClientTransport(client *testclient.ClientSet, sourcePod *corev1.Pod, sour
...
+		return hpingTCPFragV4(client, sourcePod, destinationIP, destinationPort)
+	case ingressnodefwv1alpha1.ProtocolTypeUDP:
+		if v6 {
+			return hpingUDPFragV6(client, sourcePod, destinationIP, destinationPort)
+		}
+		return hpingUDPFragV4(client, sourcePod, destinationIP, destinationPort)
+	case ingressnodefwv1alpha1.ProtocolTypeSCTP:
+		return "", "", fmt.Errorf("hping3 does not support SCTP")
+	default:
+		panic("Unsupported protocol for hping3")
```
danwinship: weird to return an error for SCTP but panic for ICMP

test/e2e/transport/transport.go
```diff
@@ -224,3 +229,62 @@ func ncClientTransport(client *testclient.ClientSet, sourcePod *corev1.Pod, sour
...
+	}
+}
+
+// IsHpingResponseSeen parses hping3 output to determine if a response was received.
+// Returns true if the target responded (packet reached it), false if dropped.
+func IsHpingResponseSeen(stdout, stderr string) bool {
+	combined := stdout + stderr
+	// hping3 prints "flags=SA" (SYN-ACK) or "flags=RA" (RST-ACK) when a response is received.
+	if strings.Contains(combined, "flags=") {
```
danwinship: This seems potentially flaky... can we distinguish based on exit status 0 vs non-0?

test/e2e/transport/transport.go
```diff
@@ -224,3 +229,62 @@ func ncClientTransport(client *testclient.ClientSet, sourcePod *corev1.Pod, sour
...
+	return false
+}
+
+func hpingTCPFragV4(client *testclient.ClientSet, sourcePod *corev1.Pod, destinationIP, dPort string) (string, string, error) {
+	return hpingClient(client, sourcePod, destinationIP, dPort, "-S", "-f", "-c", "1")
+}
+
+func hpingTCPFragV6(client *testclient.ClientSet, sourcePod *corev1.Pod, destinationIP, dPort string) (string, string, error) {
+	return hpingClient(client, sourcePod, destinationIP, dPort, "-S", "-f", "-c", "1", "-6")
+}
```
danwinship: Not sure all these helper functions actually make things clearer. It seems like it might be better to just have `HpingFragmentedConnect` assemble the hping args based on `proto`, then add a `"-6"` if `v6`, then call `exec.RunExecCommand`?

Makefile
```diff
@@ -8,7 +8,7 @@ CSV_VERSION = $(shell echo $(VERSION) | sed 's/v//')
 ifeq ($(VERSION), latest)
 CSV_VERSION := 0.0.0
 endif
-CERT_MANAGER_VERSION=v1.9.1
+CERT_MANAGER_VERSION=v1.20.2
```
danwinship, line 11: where does this change come from?

raphaelvrosa: Updated for local testing.

test/e2e/transport/transport.go
raphaelvrosa: Simplified the flags parsing and methods to a single one.

test/e2e/transport/transport.go
raphaelvrosa: Added a logic to check only returned code.

test/e2e/transport/transport.go
raphaelvrosa: Moved default to return error of unsupported protocol and removed SCTP case.

test/e2e/functional/tests/e2e.go
raphaelvrosa: Removed.

test/e2e/functional/tests/e2e.go
raphaelvrosa: Fixed.

test/e2e/functional/tests/e2e.go
raphaelvrosa: Updated comment.

test/e2e/functional/tests/e2e.go
raphaelvrosa: Removed.

test/e2e/functional/tests/e2e.go
raphaelvrosa: Indeed, removed.

bpf/ingress_node_firewall_kernel.c
raphaelvrosa: Done, added the define statements and their usage.

bpf/ingress_node_firewall_kernel.c
raphaelvrosa: Indeed, since it is stateless, better to drop all fragmented traffic.

raphaelvrosa:
@danwinship  can you please take a look again? Thanks

raphaelvrosa:
@danwinship are we moving forward with this?

anuragthehatter:
@raphaelvrosa   Fix looks good to me. Rejects any IPv4 packet with MF set or non-zero fragment offset before L4 extraction, returns DENY. My A/B confirmed the deny-port bypass is closed and allowed-port fragments are dropped.

I would say ship the fix. But test might need some changes and improvements.
- Seems test is not testing the fix? `reachabilityCheckFragmentation` has zero call sites  , seems dead code. The only fragmentation traffic that runs (inside `isConnectivitySeen`) has its result computed, logged, then discarded (return ncatResult). The suite goes green whether the fix is present or not. This must be wired in (I confirmed asserting it on the deny reachables catches the bug) or the test adds no protection against regressions.
- The PR adds hping3 to test/Dockerfile.netcat, but openshift-ci/run_e2e.sh overrides NETCAT_IMAGE=rhel8/support-tools:latest. So in CI the probe pod won't have hping3
- Not sure Ipv6 coveage is scope of this?

hack/kind-cluster.sh
```diff
@@ -3,7 +3,7 @@ set -eux
 
 DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
-KIND_IMAGE="kindest/node:v1.27.3"
+KIND_IMAGE="kindest/node:v1.36.4"
```
raphaelvrosa: Just upgraded for local tests

raphaelvrosa:
> Seems test is not testing the fix? `reachabilityCheckFragmentation` has zero call sites , seems dead code. The only fragmentation traffic that runs (inside `isConnectivitySeen`) has its result computed, logged, then discarded (return ncatResult). The suite goes green whether the fix is present or not. This must be wired in (I confirmed asserting it on the deny reachables catches the bug) or the test adds no protection against regressions.

@anuragthehatter the test is refactor, please check again. The test was not even running before, given fragmentation was not being exercised (packets were assembled in NIC, so a large packet enforces fragmentation). The test runs locally fine. Dan mentioned to add ipv6 support for this change, so ipv6 fragmentation is also verified in this bug fix, and tests are included/refactored.

bpf/ingress_node_firewall_kernel.c
```diff
@@ -107,23 +109,52 @@ ip_extract_l4info(void *data, void *dataEnd, __u8 *proto, __u16 *dstPort,
...
+      return L4_TRUNCATED;
     }
     *proto = iph->nexthdr;
+
+    /*
+     * Walk the IPv6 extension header chain looking for a Fragment header.
+     * INF cannot reassemble, so any fragment is denied, same as IPv4 above.
+     */
+#pragma clang loop unroll(full)
+    for (int i = 0; i < MAX_IPV6_EXT_HDRS; ++i) {
```
danwinship: This lets you bypass the check by inserting extra headers. We should fail in that case.

raphaelvrosa: Added a check for the number of extension headers, returning a specific error code for it.
Didn't check that in e2e tests, is that fine?

PR #754, opened 2026-07-22
OCPBUGS-99202: Prevent daemon panic on out-of-range rule order

smulje:
## Summary
- Cap `IngressNodeFirewallProtocolRule.order` at 99 in the CRD schema (covers both `IngressNodeFirewall` and `IngressNodeFirewallNodeState`)
- Reject out-of-range `order` in the admission webhook
- Validate `order` in the daemon before using it as an eBPF rules-array index, so a crafted NodeState cannot panic every node's reconciler into CrashLoopBackOff

`order` is consumed as a raw index into a fixed `[100]BpfRuleTypeSt` array (index 0 reserved). Previously only `minimum: 1` was enforced, and the webhook only checked rule count / uniqueness — any `order >= 100` (up to `uint32` max) panicked the daemon cluster-wide.

## Test plan
- [x] `go test ./pkg/failsaferules/`
- [ ] Webhook suite: rejects `order == 100` and `order == ^uint32(0)`; accepts `order == 99`
- [x] Apply an `IngressNodeFirewall` with `order: 100` and confirm API/webhook rejection
- [x] Apply an `IngressNodeFirewallNodeState` with `order: 100` and confirm CRD rejection (`should be less than or equal to 99`)
- [x] Confirm `order: 99` is accepted and daemons attach XDP without panic

## Cluster verification (OCP 4.21)
Reproduced the panic on stock 4.21 (`index out of range [100] with length 100`), then verified the fix with custom images built from this branch:

- Operator: `quay.io/smulje/ingress-node-firewall-operator:order-oob-fix`
- Daemon: `quay.io/smulje/ingress-node-firewall-daemon:order-oob-fix`

Results with the custom images:
- `order: 100` / `order: 4294967295` rejected by webhook
`An error occurred
Error "Invalid value: 100: spec.ingress[0].rules[0].order in body should be less than or equal to 99" for field "spec.ingress[0].rules[0].order".`
- `order: 99` applied successfully (`Synchronized`), daemons remained Running
- Direct `IngressNodeFirewallNodeState` write with `order: 100` rejected by updated CRD schema

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to reject firewall rule orders outside the supported range of 1–99.
  * Prevented invalid rule orders from causing runtime indexing errors.

* **Documentation**
  * Clarified that rule orders must be unique and that order 0 is reserved for internal catch-all handling.
  * Updated CRD schemas with the maximum order validation.

* **Tests**
  * Added coverage for minimum, maximum, and out-of-range rule order values.

### DIFF
--- a/api/v1alpha1/ingressnodefirewall_types.go
+++ b/api/v1alpha1/ingressnodefirewall_types.go
@@ -90,10 +90,11 @@ type IngressNodeProtocolConfig struct {
 // IngressNodeFirewallProtocolRule defines an ingress node firewall rule per protocol.
 type IngressNodeFirewallProtocolRule struct {
 	// order defines the order of execution of ingress firewall rules.
-	// The minimum order value is 1 and the values must be unique.
-	// + index 0 is used internally as catch all for unclassified packets matching the same sourceCIDR.
+	// The minimum order value is 1, the maximum is 99, and the values must be unique.
+	// Index 0 is used internally as catch all for unclassified packets matching the same sourceCIDR.
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Minimum:=1
+	// +kubebuilder:validation:Maximum:=99
 	Order uint32 `json:"order"`
 
 	// protocolConfig is a discriminated union of a protocol's specific configuration for TCP, UDP, SCTP, ICMP and ICMPv6.
--- a/pkg/ebpf/ingress_node_firewall_loader.go
+++ b/pkg/ebpf/ingress_node_firewall_loader.go
@@ -521,6 +521,13 @@ func (infc *IngNodeFwController) makeIngressFwRulesMap(
 	for _, rule := range ingFirewallConfig.FirewallProtocolRules {
 		rule := rule
 		idx := rule.Order
+		// Order is used as a raw index into the fixed-size eBPF rules array.
+		// Bound-check against the array length so a crafted IngressNodeFirewallNodeState
+		// (which has no admission webhook) cannot panic the daemon.
+		if idx == 0 || int(idx) >= len(rules.Rules) {
+			return keys, rules, fmt.Errorf("rule order %d out of range; must be between 1 and %d",
+				idx, len(rules.Rules)-1)
+		}
 		rules.Rules[idx].RuleId = rule.Order
 		switch rule.ProtocolConfig.Protocol {
 		case ingressnodefwiov1alpha1.ProtocolTypeTCP:
--- a/pkg/failsaferules/failsaferules.go
+++ b/pkg/failsaferules/failsaferules.go
@@ -1,6 +1,6 @@
 package failsaferules
 
-var MAX_INGRESS_RULES = 100
+const MAX_INGRESS_RULES = 100
 
 type TransportProtoFailSafeRule struct {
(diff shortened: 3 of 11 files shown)


api/v1alpha1/ingressnodefirewall_types.go
```diff
@@ -90,10 +90,12 @@ type IngressNodeProtocolConfig struct {
 // IngressNodeFirewallProtocolRule defines an ingress node firewall rule per protocol.
 type IngressNodeFirewallProtocolRule struct {
 	// order defines the order of execution of ingress firewall rules.
-	// The minimum order value is 1 and the values must be unique.
-	// + index 0 is used internally as catch all for unclassified packets matching the same sourceCIDR.
+	// The minimum order value is 1, the maximum is 99, and the values must be unique.
+	// Index 0 is used internally as catch all for unclassified packets matching the same sourceCIDR.
+	// Order is consumed as an index into a fixed-size eBPF rules array of length 100.
```
danwinship: That last sentence is not part of the _API_; it's an implementation detail, which could be changed in the future, so it shouldn't be documented here.

pkg/ebpf/ingress_node_firewall_loader.go
```diff
@@ -521,6 +522,12 @@ func (infc *IngNodeFwController) makeIngressFwRulesMap(
 	for _, rule := range ingFirewallConfig.FirewallProtocolRules {
 		rule := rule
 		idx := rule.Order
+		// Order is used as a raw index into the fixed-size eBPF rules array.
+		// Reject out-of-range values here so a crafted IngressNodeFirewallNodeState
+		// (which has no admission webhook) cannot panic the daemon.
+		if err := failsaferules.ValidateRuleOrder(idx); err != nil {
+			return keys, rules, err
+		}
 		rules.Rules[idx].RuleId = rule.Order
```
danwinship, line 531: It would be more robust to just check `idx` against `len(rules.Rules)`, so that even if `ValidateRuleOrder` was wrong, it wouldn't crash

pkg/failsaferules/failsaferules.go
```diff
@@ -1,6 +1,25 @@
...
+	MAX_INGRESS_RULES = 100
+	// MaxRuleOrder is the highest Order value that can be used as an index into the
+	// eBPF rules array without panicking.
+	MaxRuleOrder = MAX_INGRESS_RULES - 1
+)
+
+// ValidateRuleOrder reports whether order can be used as an index into the
+// fixed-size eBPF rules array. Index 0 is reserved; values >= MAX_INGRESS_RULES
+// are out of bounds.
+func ValidateRuleOrder(order uint32) error {
```
danwinship: `failsaferules` is about blocking valid-but-dangerous rules, like "deny all access to the apiserver". In this case we are checking for an invalid rule, so it doesn't belong in this package.

(But since you don't need to call this from `ingress_node_firewall_loader.go` now, the check can just be inlined into `validateRules` below, like the other checks.)

danwinship (commented):
a few comments

smulje:
/cherry-pick release-4.22 release-4.21 release-4.20 release-4.19 release-4.18

PR #760, opened 2026-07-30
OCPBUGS-96585 OCPBUGS-96584: CVE-2026-27136 openshift4/ingress-node-firewall-rhel9-operator: golang.org/x/net/html

dasanind:
FIXES: [OCPBUGS-96585](https://redhat.atlassian.net/browse/OCPBUGS-96585), [OCPBUGS-96584](https://redhat.atlassian.net/browse/OCPBUGS-96584)

- Bump golang.org/x/net from v0.37.0 to v0.55.0
- Required transitive bumps
   - golang.org/x/crypto from v0.36.0 to v0.51.0 ==> Required by x/net v0.55.0
   - golang.org/x/sys from v0.32.0 to v0.45.0 ==> Required by x/net v0.55.0
   - golang.org/x/term from v0.30.0 to v0.43.0 ==> Required by x/net v0.55.0
   - golang.org/x/text from v0.23.0 to v0.37.0 ==> Required by x/net v0.55.0
   - golang.org/x/sync from v0.12.0 to v0.20.0 ==> Required by x/net v0.55.0
   
Signed-off-by: [anidas@redhat.com](mailto:anidas@redhat.com)


MrSanketkumar:
Hi @dasanind, thanks for raising this PR. I wanted to clarify the intent here I've done a full analysis (govulncheck + 4-algorithm call graph + vendor inspection + manual review) and confirmed that the golang.org/x/net/html vulnerable functions **(Parse, ParseFragment, Render) are never called in our codebase.** The dependency chain is test-only **(k8sreporter → sanitize.BaseName() which is pure string manipulation, no HTML parsing).** So from a CVE-2026-27136 perspective, we're not vulnerable and the trackers could be closed as "Not a Bug." was this PR raised for compliance purposes rather than exploitability? If it's compliance-driven, we'd need to apply the same bump across all tracked versions (4.22 through 4.12), not just this one. Would appreciate your clarification so we can align on the right approach.

dasanind:
> Hi @dasanind, thanks for raising this PR. I wanted to clarify the intent here I've done a full analysis (govulncheck + 4-algorithm call graph + vendor inspection + manual review) and confirmed that the golang.org/x/net/html vulnerable functions **(Parse, ParseFragment, Render) are never called in our codebase.** The dependency chain is test-only **(k8sreporter → sanitize.BaseName() which is pure string manipulation, no HTML parsing).** So from a [CVE-2026-27136](https://github.com/advisories/GHSA-m9x8-m34x-fj9q) perspective, we're not vulnerable and the trackers could be closed as "Not a Bug." was this PR raised for compliance purposes rather than exploitability? If it's compliance-driven, we'd need to apply the same bump across all tracked versions (4.22 through 4.12), not just this one. Would appreciate your clarification so we can align on the right approach.

Hi @MrSanketkumar , That's correct this vulnerability is not in the execution path. The only code path to golang.org/x/net/html is through test/e2e/k8sreporter/reporter.go →kennygrant/sanitize → sanitize.BaseName(), which is regex-only and never calls the HTML parser. The fix is compliance-driven only. There are similar tickets raised for 4.22 and 4.21. I didn't fix it there as it needed the Go version upgrade. Do you suggest not to fix this for compliance purposes?

MrSanketkumar:
> Hi @MrSanketkumar , That's correct this vulnerability is not in the execution path. The only code path to golang.org/x/net/html is through test/e2e/k8sreporter/reporter.go →kennygrant/sanitize → sanitize.BaseName(), which is regex-only and never calls the HTML parser. The fix is compliance-driven only. There are similar tickets raised for 4.22 and 4.21. I didn't fix it there as it needed the Go version upgrade. Do you suggest not to fix this for compliance p

I mean, I don’t have any issue fixing CVEs for compliance. If the maintainer wants us to fix it, we’re happy to do so.

Regarding the Go bump, We (Sustaining team) should have a fork/patch available that addresses this CVE, so we may be able to use that approach and avoid the Go bump if needed. However, I’m not sure how long it will take to get the fork/patch ready.


So, if the maintainer asks us to fix it, we can take care of it. Otherwise, I think we should close the tracker. That’s just my opinion, because all of these trackers are in the black bucket, and keeping them open for a long time doesn’t really add much value.

dasanind:
> > Hi @MrSanketkumar , That's correct this vulnerability is not in the execution path. The only code path to golang.org/x/net/html is through test/e2e/k8sreporter/reporter.go →kennygrant/sanitize → sanitize.BaseName(), which is regex-only and never calls the HTML parser. The fix is compliance-driven only. There are similar tickets raised for 4.22 and 4.21. I didn't fix it there as it needed the Go version upgrade. Do you suggest not to fix this for compliance p
> 
> I mean, I don’t have any issue fixing CVEs for compliance. If the maintainer wants us to fix it, we’re happy to do so.
> 
> Regarding the Go bump, We (Sustaining team) should have a fork/patch available that addresses this CVE, so we may be able to use that approach and avoid the Go bump if needed. However, I’m not sure how long it will take to get the fork/patch ready.
> 
> So, if the maintainer asks us to fix it, we can take care of it. Otherwise, I think we should close the tracker. That’s just my opinion, because all of these trackers are in the black bucket, and keeping them open for a long time doesn’t really add much value.

@MrSanketkumar As per the above discussion, closing the PR.

PR #771, opened 2026-08-18
OCPBUGS-105459: Remove kube-rbac-proxy sidecar and enable native TLS metrics in daemon

asood-rh:
Fixes: https://redhat.atlassian.net/browse/OCPBUGS-105459

Replace the kube-rbac-proxy sidecar container with controller-runtime's built-in TLS serving for the daemon metrics endpoint. The daemon now handles TLS termination, authentication, and authorization directly using the cluster's TLS profile configuration.

- Add TLS flags and secure metrics serving to cmd/daemon/daemon.go
- Remove kube-rbac-proxy container from daemon daemonset manifest
- Remove KUBE_RBAC_PROXY_IMAGE references from controller, main.go, and CSV
- Delete orphaned config/default/manager_auth_proxy_patch.yaml
- Daemon serves HTTPS metrics on :9301 with certs from the metrics Service
- Services RBAC in CSV for controller metrics Service reconcile

As part https://redhat.atlassian.net/browse/CORENET-7108 implementation for TLS was done for controller that involved removal of kube-rbac-proxy sidecar. The sidecar was not removed for  daemonset that resulted in causing the bug where ingress node firewall operator could be installed on  ppc64le but daemonset could not be created.

The changes are only for ingress node firewall daemonset as controller already has TLS support.

**- How to verify it**
No additional tests added but existing tests modified.

make test
make test-race
go test ./controllers/... -count=1

Completes the TLS support for ingress node firewall operator.

## New Features
- Metrics are now served securely over HTTPS on port 39401 with configurable TLS settings.
- Added daemon liveness and readiness health checks.
- Added permission to retrieve metrics through authenticated service-account requests.

## Improvements
- Simplified metrics access by removing the separate proxy component and related configuration.
- Metrics certificates are now required and mounted read-only.
- Removed the proxy image startup requirement.

## Documentation & Tests
- Updated metrics examples and end-to-end tests for HTTPS and service-account authentication.


test/e2e/functional/tests/e2e.go
```diff
@@ -1368,7 +1368,7 @@ var _ = Describe("Ingress Node Firewall", func() {
 			var stdOut, stdError string
 			var metrics testutil.Metrics
 			err = wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 30*time.Second, true, func(ctx context.Context) (done bool, err error) {
-				stdOut, stdError, err = exec.RunExecCommand(testclient.Client, daemonSetPod, "/usr/bin/curl", "127.0.0.1:39401/metrics")
+				stdOut, stdError, err = exec.RunExecCommand(testclient.Client, daemonSetPod, "/usr/bin/curl", "-k", "https://127.0.0.1:39401/metrics")
```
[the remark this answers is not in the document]
asood-rh: @coderabbitai Resolve after review if it is fixed.

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -26,6 +26,10 @@ spec:
       containers:
         - name: daemon
           image: '{{.Image}}'
+          args:
+            - --metrics-bind-address=:39401
+            - --metrics-cert-dir=/etc/pki/tls/metrics-certs
+            - --metrics-secure=true
```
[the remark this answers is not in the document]
asood-rh: @coderabbitai Resolve after review if it is fixed, except adding network policy and privileged SCC

asood-rh:
@danwinship @knobunc @tpantelis Please help review

cmd/daemon/daemon.go
```diff
@@ -80,9 +99,44 @@ func main() {
...
+	if err != nil {
+		setupLog.Error(err, "unable to create client")
+		os.Exit(1)
+	}
+
+	tlsProfile, err := inftls.GetProfileInfo(logr.NewContext(ctx, setupLog), tlsProfileClient)
+	if err != nil {
+		setupLog.Error(err, "unable to get TLS profile options")
+		os.Exit(1)
+	}
```
tpantelis, line 108: The daemon is an operand of the ingress node firewall operator so I think we should pass the TLS flags as args as was done with the kube-rbac-proxy sidecar, ie

```
{{- if .TLSMinVersion }}
   --tls-min-version={{.TLSMinVersion}} \
{{- end }}
{{- if .TLSCipherSuites }}
    --tls-cipher-suites={{.TLSCipherSuites}} \
{{- else }}
    --tls-cipher-suites=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256 \
{{- end }}
```

It won't need the `apiservers` RBAC perm either.

bindata/manifests/daemon/daemonset.yaml
tpantelis: To address the issue noted by CodeRabbit where the `Secret` is absent at startup, we could set `optional` to false here. Then the pod won't start until the `Secret` exists. This is simpler than the `newDeferredCertProvider`. I believe the comment above is incorrect (or perhaps used to be the behavior), ie once the Openshift `service-ca` operator sees the serving cert annotation, it creates the Secret immediately. Also I did the same with the ingress node firewall operator and it's been working fine with `optional: false`. It might delay the initial startup of the pod but that should be negligible.

tpantelis (changes requested):

cmd/daemon/daemon.go
```diff
@@ -80,9 +94,45 @@ func main() {
...
+	if tlsCipherSuites != "" {
+		ids, err := parseCipherSuites(tlsCipherSuites)
+		if err != nil {
+			setupLog.Error(err, "invalid TLS cipher suites")
+			os.Exit(1)
+		}
+		tlsOpts = append(tlsOpts, func(c *tls.Config) {
+			c.CipherSuites = ids
+		})
+	}
```
tpantelis, line 119: There's utilities for parsing TLS flags suitable for `tls.Config`:

```Go
import cliflag "k8s.io/component-base/cli/flag"
...

minVersionID, err := cliflag.TLSVersion(tlsMinVersion)
if err != nil {
    return nil, err
}

cipherSuiteIDs, err := cliflag.TLSCipherSuites(strings.Split(tlsCipherSuites, ","))
if err != nil {
    return nil, err
}

tlsOpts := []func(*tls.Config){
    func(c *tls.Config) {
        c.MinVersion = minVersionID
        c.CipherSuites = cipherSuiteIDs
    },
}
```

tpantelis (changes requested):

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -26,6 +26,13 @@ spec:
       containers:
         - name: daemon
           image: '{{.Image}}'
+          command: ["/usr/bin/daemon"]
+          args:
+            - --metrics-bind-address=:39401
+            - --metrics-cert-dir=/etc/pki/tls/metrics-certs
+            - --metrics-secure=true
+            - '--tls-min-version={{.TLSMinVersion}}'
+            - '--tls-cipher-suites={{default "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384" .TLSCipherSuites}}'
```
tpantelis: Why the surrounding single quotes?

asood-rh: Single quotes to avoid breaking YAML parsing due to embedded double quotes in the curl braces. For .TLSMinVersion for consistency purpose. Tested it out parsing, so will remove them.

asood-rh: Fixed by removing single quotes.

cmd/daemon/daemon.go
asood-rh: Using k8s utility for parsing TLS flags.

cmd/daemon/daemon.go
asood-rh: Done

bindata/manifests/daemon/daemonset.yaml
asood-rh: Done

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -57,67 +86,11 @@ spec:
...
-              cpu: 10m
-              memory: 20Mi
-          terminationMessagePolicy: FallbackToLogsOnError
-          volumeMounts:
-            - name: ingress-node-firewall-daemon-metrics-certs
-              mountPath: /etc/pki/tls/metrics-certs
-              readOnly: True
       volumes:
-        # Must be optional because the ingress-node-firewall-daemon-metrics-certs is a service serving
-        # certificate and those cannot be generated without the daemon running
```
tpantelis, line 116: Can you remove these comment as they no longer apply?

asood-rh: It is already removed when optional was set to false.

asood-rh:
@tpantelis E2E tests passed but failing at step not related to these changes.
<pre>
  1. Extension test (infw-extension step):
  error: resource mapping not found for name: "infw-extensions"
  no matches for kind "TestExtensionAdmission" in version "testextension.redhat.io/v1"
  ensure CRDs are installed first
    - Trying to create a TestExtensionAdmission CR
    - The CRD doesn't exist on the cluster
    - This is a separate feature test, not part of core ingress-node-firewall functionality
</pre>
E2E tests passing
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_ingress-node-firewall/771/pull-ci-openshift-ingress-node-firewall-master-ingress-node-firewall-e2e-metal-ipi/2090413447853379584/artifacts/ingress-node-firewall-e2e-metal-ipi/baremetalds-ingress-node-firewall-e2e-test/build-log.txt

@anuragthehatter Did any related to OTE merged in the repo?

tpantelis (approved):

asood-rh:
@danwinship @knobunc Could you help merge this PR as this needs to back ported to 5.0 otherwise will be blocked for testing ingress node firewall.  This is release blocker bug fix. Thanks!

prb112:
Thanks @asood-rh 

/lgtm

tpantelis:
It seems the root cause issue is that `kube-rbac-proxy` was mistakenly removed from `manifests/stable/image-references` by https://github.com/openshift/ingress-node-firewall/pull/720/changes/100a5a2c413eb956eb6d86f8a448041d37a65a89 since the ingress firewall node operator no longer uses it.  But the daemonset operand still uses it. So adding it back to the `manifests/stable/image-references` should fix this issue.

asood-rh:
> It seems the root cause issue is that `kube-rbac-proxy` was mistakenly removed from `manifests/stable/image-references` by [100a5a2](https://github.com/openshift/ingress-node-firewall/commit/100a5a2c413eb956eb6d86f8a448041d37a65a89) since the ingress firewall node operator no longer uses it. But the daemonset operand still uses it. So adding it back to the `manifests/stable/image-references` should fix this issue.

Alternatively it is fine to remove the side car for the following reasons:-
<pre>
  1. Less overhead: No sidecar container (saves memory, CPU, pod scheduling complexity)
  2. Simpler deployment: One container vs two
  3. Direct control: TLS config in the code, not in proxy config
  4. Maintained by controller-runtime: Red Hat actively maintains this path
  5. Same security: filters.WithAuthenticationAndAuthorization provides identical bearer token + RBAC enforcement
  6. Future-proof: This is the direction OpenShift is moving

</pre>

tpantelis:
> > It seems the root cause issue is that `kube-rbac-proxy` was mistakenly removed from `manifests/stable/image-references` by [100a5a2](https://github.com/openshift/ingress-node-firewall/commit/100a5a2c413eb956eb6d86f8a448041d37a65a89) since the ingress firewall node operator no longer uses it. But the daemonset operand still uses it. So adding it back to the `manifests/stable/image-references` should fix this issue.
> 
> Alternatively it is fine to remove the side car for the following reasons:-
> 

That's fine for 5.1 but perhaps we should add the image-reference back and keep the `kube-rbac-proxy` for 5.0 to reduce risk.

asood-rh:
/verified by CI

No additional tests needed for the code change.

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_ingress-node-firewall/771/pull-ci-openshift-ingress-node-firewall-master-ingress-node-firewall-e2e-metal-ipi/2090413447853379584 

<pre>
  
  [OCPFeatureGate:TLSAdherence][Serial] TLS Profile Compliance - should verify ingress-node-firewall TLS compliance across all profiles
     
  This test validates the TLS race condition fix by checking all three TLS profile scenarios:

  1. Modern + LegacyAdheringComponentsOnly (TLS 1.2 and 1.3 both work)
  2. Modern + StrictAllComponents (TLS 1.3 only, TLS 1.2 rejected)
  3. Intermediate + StrictAllComponents (TLS 1.2 and 1.3 both work) ← This is the critical test case

</pre>

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -41,10 +48,32 @@ spec:
               value: '{{.Debug}}'
             - name: EBPF_MANAGEMENT_MODE
               value: '{{.EBPFProgramManagerMode}}'
+          ports:
+            - containerPort: 39401
+              name: https
```
danwinship:
```suggestion
              name: metrics
```

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -41,10 +48,32 @@ spec:
...
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: 39400
+              host: 127.0.0.1
+            initialDelaySeconds: 15
+            periodSeconds: 20
+            timeoutSeconds: 3
+          readinessProbe:
```
danwinship: where did the liveness and readiness probes come from / why are they being added as part of this PR?

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -57,67 +86,11 @@ spec:
...
-              mountPath: /etc/pki/tls/metrics-certs
-              readOnly: True
       volumes:
-        # Must be optional because the ingress-node-firewall-daemon-metrics-certs is a service serving
-        # certificate and those cannot be generated without the daemon running
         - name: ingress-node-firewall-daemon-metrics-certs
           secret:
             secretName: ingress-node-firewall-daemon-metrics-certs
-            optional: true
+            optional: false
```
danwinship: you can just remove this; not-optional is the default

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -41,10 +48,32 @@ spec:
               value: '{{.Debug}}'
             - name: EBPF_MANAGEMENT_MODE
               value: '{{.EBPFProgramManagerMode}}'
+          ports:
+            - containerPort: 39401
```
danwinship: Before, there was an http metrics listener on `127.0.0.1:39401`, which was re-exported via https on `:9301` by the kube-rbac-proxy. It's the second port that you want to preserve. The higher-numbered port may be firewalled and inaccessible to other nodes, but we require ports 9000-9999 to be unfirewalled.

(Note that the 39xxx and 9xxx ports don't match here: 39**4**01 vs 9**3**01. That's probably a mistake but it doesn't matter any more. Stick with 9301.)

cmd/daemon/daemon.go
```diff
@@ -50,9 +54,19 @@ func init() {
...
+	var tlsCipherSuites string
 	// We are host networked, we set default to loopback by default
 	flag.StringVar(&probeAddr, "health-probe-bind-address", "127.0.0.1:39400", "The address the probe endpoint binds to.")
-	flag.StringVar(&metricsAddr, "metrics-bind-address", "127.0.0.1:39401", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":39401", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsCertDir, "metrics-cert-dir", "/etc/pki/tls/metrics-certs", "Directory containing TLS certificates for metrics endpoint.")
+	flag.BoolVar(&secureMetrics, "metrics-secure", true, "If the metrics endpoint should be served securely.")
+	flag.BoolVar(&enableHTTP2, "enable-http2", false, "If HTTP/2 should be enabled for the metrics server.")
+	flag.StringVar(&tlsMinVersion, "tls-min-version", "", "Minimum TLS version for the metrics endpoint (e.g. VersionTLS12, VersionTLS13).")
+	flag.StringVar(&tlsCipherSuites, "tls-cipher-suites", "", "Comma-separated list of TLS cipher suites for the metrics endpoint.")
```
danwinship: if these flags only affect metrics they should have `metrics-` in their names like `metrics-secure`

controllers/suite_test.go
```diff
@@ -147,7 +146,6 @@ var _ = BeforeSuite(func() {
 }, 60)
 
 var _ = AfterEach(func() {
-	// Reset the TLS profile spec to prevent test cross-contamination
 	tlsProfileSpec.Store((*configv1.TLSProfileSpec)(nil))
```
danwinship, line 150: I'd leave this comment. It seems useful.

openshift-ci/deploy_ingress_node_firewall.sh
```diff
@@ -165,8 +165,8 @@ sleep_time=10
 max_iterations=72 # results in 12 minutes timeout
 until $ds_ready
 do
-  desired_ds_num=$(oc get ds -n "$NAMESPACE" ingress-node-firewall-daemon -o jsonpath="{.status.desiredNumberScheduled}")
-  ready_ds_num=$(oc get ds -n "$NAMESPACE" ingress-node-firewall-daemon -o jsonpath="{.status.numberReady}")
+  desired_ds_num=$(oc get ds -n "$NAMESPACE" ingress-node-firewall-daemon -o jsonpath="{.status.desiredNumberScheduled}" 2>/dev/null || echo "0")
+  ready_ds_num=$(oc get ds -n "$NAMESPACE" ingress-node-firewall-daemon -o jsonpath="{.status.numberReady}" 2>/dev/null || echo "0")
```
danwinship, line 169: what is this for?

danwinship (commented):
Please squash the commits. The PR should just reflect the changes that are being made, not the chain of events that led there.

bindata/manifests/daemon/daemonset.yaml
asood-rh: They were not there but were added. Without them 
<pre>
 1. No automatic failure detection - If the daemon deadlocks or the health endpoint stops responding, the pod stays "Running" forever
  2. No automatic restarts - Without liveness probes, failed pods won't be restarted by kubelet
  3. Immediate traffic during rollouts - Without readiness probes, pods are considered ready the instant they start, even if initialization isn't complete
  4. Manual intervention required - When things break, you have to manually identify and delete bad pods

</pre>

openshift-ci/deploy_ingress_node_firewall.sh
asood-rh: The 2>/dev/null || echo "0" pattern ensures the script doesn't crash or spam errors during the normal "resource doesn't exist yet" phase.

bindata/manifests/daemon/daemonset.yaml
danwinship: That sounds like AI and is not correct in this context. Please remove the probes.

README.md
```diff
@@ -312,9 +312,10 @@ within the node daemons themselves:
 ```sh
 kubectl exec -n ${OPERATOR_NAMESPACE} -it ${NODE_DAEMON_NAME} sh
 ```
-2. Retrieve the Prometheus formatted metrics
+2. Retrieve the Prometheus-formatted metrics (requires a bearer token with GET permission on the `/metrics` non-resource URL, granted to the daemon service account via `ingress-node-firewall-daemon-clusterrole-extra`)
 ```sh
-Curl 127.0.0.1:39401/metrics
+TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
+curl -k -H "Authorization: Bearer ${TOKEN}" https://127.0.0.1:39401/metrics
```
danwinship:
```suggestion
curl -k -H "Authorization: Bearer ${TOKEN}" https://127.0.0.1:9301/metrics
```

danwinship:
looks good after you remove the added probes

README.md
asood-rh: Just updated it and the commit message.

asood-rh:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_ingress-node-firewall/771/pull-ci-openshift-ingress-node-firewall-master-ingress-node-firewall-e2e-metal-ipi/2092649730499678208

--- PASS: TestE2E (3096.64s)
PASS
ok  	github.com/openshift/ingress-node-firewall/test/e2e/functional	3096.696s

PR #770, opened 2026-08-18
OCPBUGS-111895: Add missing RBAC permissions for networking.k8s.io/ne…

weliang1:
The ingress-node-firewall-operator CSV was missing required RBAC permissions for networking.k8s.io/networkpolicies, causing the operator bundle to fail the Conforma policy check 'olm.required_network_policy_rbac_for_operands'.

This fix adds the required permissions by:
- Adding kubebuilder RBAC marker to ingressnodefirewall_controller.go
- Regenerating RBAC manifests with 'make manifests'
- Updating all CSV files in bundle/ and manifests/stable/

## Summary by CodeRabbit

* **New Features**
  * Added a default network policy for the ingress node firewall daemon.
  * The policy blocks incoming traffic while allowing outgoing traffic.
  * The operator now manages network policies throughout their lifecycle.
* **Bug Fixes**
  * Improved deployment handling to ensure rendered daemon resources are applied correctly.

PR #784, opened 2026-08-27
OCPBUGS-104598: Add missing RBAC permissions for NetworkPolicy

weliang1:
Backported from PR #770 (OCPBUGS-111895).

The ingress-node-firewall-operator CSV was missing required RBAC permissions for networking.k8s.io/networkpolicies, causing the operator bundle to fail the Conforma policy check
'olm.required_network_policy_rbac_for_operands'.

This fix adds the required permissions by:
- Adding a NetworkPolicy manifest for the daemon
- Adding kubebuilder RBAC marker to ingressnodefirewall_controller.go
- Updating RBAC manifests in config/rbac/role.yaml
- Updating all CSV files in bundle/ and manifests/stable/

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -143,3 +143,15 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - networking.k8s.io
+  resources:
+  - networkpolicies
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
--- a/controllers/ingressnodefirewallconfig_controller.go
+++ b/controllers/ingressnodefirewallconfig_controller.go
@@ -34,6 +34,7 @@ import (
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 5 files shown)


config/rbac/role.yaml
```diff
@@ -12,18 +12,6 @@ rules:
   - get
   - list
   - watch
-- apiGroups:
-  - apps
-  resources:
-  - daemonsets
```
danwinship, line 30: this file needs to be regenerated

controllers/ingressnodefirewallconfig_controller.go
```diff
@@ -267,6 +270,18 @@ func (r *IngressNodeFirewallConfigReconciler) syncIngressNodeFwConfigResources(c
 			if err := apply.ApplyObject(ctx, r.Client, obj); err != nil {
 				return errors.Wrapf(err, "could not apply (%s) %s", obj.GroupVersionKind(), obj.GetName())
 			}
+
+			if err := apply.ApplyObject(ctx, r.Client, obj); err != nil {
+				return errors.Wrapf(err, "could not apply (%s) %s", obj.GroupVersionKind(), obj.GetName())
+			}
```
danwinship: diffing this against 782, it seems that these lines shouldn't be here (indeed, it's clearly just an accidental copy of the immediately-preceding lines)

weliang1: Just remove those three lines.

